PR #239, opened 2016-04-19
(fixup) remove incorrect uses of 'nfs'

rji:
In PR #210, I incorrectly referred to the backing storage as NFS. On
Azure, which is used in this tutorial, the proper protocol is actually
CIFS.

This commit corrects the incorrect usage, and perhaps more
appropriately, generalizes the technology as an "external volume"

/cc @mhausenblas, @pyronicide


ssk2:
lgtm!

PR #242, opened 2016-04-20
pull nginx tutorial

joel-hamill:
per @gabrielhartmann this is broken


ssk2:
@gabrielhartmann - what was the issue? I just updated the Universe package for nginx, so wondering if I broke it!

gabrielhartmann:
@ssk2 The issue is that you can't see the nginx instance from outside the cluster.  You can't see it either through admin router or through the public IP.

ssk2:
Okay, I'll look into that - and will fix this tutorial. (We should try to fix the underlying issues instead of removing docs!)

joel-hamill:
@ssk2 the docs should be removed until it is fixed though. it reflects poorly on us. as soon as the code is fixed, let's push the doc back out.

mohitsoni:
@gabrielhartmann That's not true, I just ran through all the tutorial steps and it worked for me as it's documented in the tutorial. You can access nginx after you install it via `<dcos-hostname>/service/nginx` URL. I wonder how you are trying to access the cluster. 

Also, you need to remember the it takes a while before adminrouter starts routing stuff to NGINX due to the delay caused by mesos-dns.

cc: @ssk2 @joel-hamill @pyronicide

ssk2:
Should we revert this PR?

mohitsoni:
Did somebody tried and confirmed that the tutorial didn't work, before reverting it ?

joel-hamill:
@mohitsoni @gabrielhartmann i have not. one thought is, which version of DC/OS are you trying? Remember that DC/OS 1.7 is EA, DC/OS 1.6 is GA. The docs on docs.mesosphere are 1.6 GA, docs.io are 1.7 EA.

PR #258, opened 2016-04-25
Follow the redirect with curl

surdy:
We need `-L` since `latest` is a symlink.

Fixes [DCOS-40](https://dcosjira.atlassian.net/browse/DCOS-40)


pyronicide:
@surdy This should be hardcoded to 1.7 instead of usage. Can you update?

surdy:
Yeah, that makes more sense. Updated

PR #261, opened 2016-04-27
Use full path to ip

surdy:
Use full path to `ip` . If you don't some things might fail. Kafka brokers failed in one instance.


surdy:
On a second look this is not needed due to `export` line

PR #262, opened 2016-04-28
Slave agent fix

joel-hamill:


mhausenblas:
Thanks @joel-hamill — now, I take it you did a sort of grep over the docs? Did you check if certain commands, for example accessing `leader.mesos:5050/agents` really work?

joel-hamill:
@mhausenblas good catch. that wont work. i have fixed, and there shouldn't be any others like that.

mhausenblas:
Perfect, thanks! We might need to do a second pass one day to make sure that also screen shots are covered but this LGTM now—merging!

PR #266, opened 2016-04-29
Updated the external volume support link.

jay-lau:
The link for external volume is not right, we should update it to docker volume driver design document link.


mhausenblas:
Thanks for pointing this out @jay-lau!

PR #267, opened 2016-04-29
Update auth-api.md

mitel:
Added a note on how to find the authtoken to avoid confusion with the one generated in the browser window.

PR #274, opened 2016-05-02
Demonstrate to users how to retrieve auth token

jgarcia-mesosphere:
Building on #267 to personalize the instructions and show bash expansion.


mhausenblas:
Perfect, thanks!

PR #281, opened 2016-05-03
Remove build status badge.

ssk2:
There's no build for this project in Travis anymore (it's built as part of the DC/OS Website build).

cc @pyronicide


karlkfi:
The build in dcos-docs was for fast feedback so that authors could fix their PRs immediately instead of waiting for a merge. Why was that removed?

ssk2:
It was linking to a Travis CI build that does not apparently exist. Is
there new build we should be showing status for?

karlkfi:
When the projects got made public they got moved automatically to .org instead of .com
Apparently this one never got reactivated. I just activated it: https://travis-ci.org/dcos/dcos-docs

@ssk2: Can you make a PR to re-add the status badge?

If you want you can make an issue to move it to Velocity, but it's pretty trivial.

PR #291, opened 2016-05-09
Change pointer for packet-terraform

jgarcia-mesosphere:
After gifting the `packet-terraform` repo to the DC/OS team, the next step is to change the docs.


mhausenblas:
LGTM and thanks a lot @jgarcia-mesosphere!

PR #70, opened 2016-04-14
Change config file examples to require users fill in resolvers

sargun:

PR #293, opened 2016-05-10
Change config file examples to require users fill in resolvers

jgarcia-mesosphere:
As in #70


mhausenblas:
Perfect, thanks @jgarcia-mesosphere!

PR #296, opened 2016-05-10
added info about google dns alternatives

joel-hamill:
@mhausenblas how does this look?


mhausenblas:
Awesome, thank you @joel-hamill!

PR #302, opened 2016-05-12
fix

joel-hamill:


joel-hamill:
@sascala regarding "via" vs "using" - this is an old IBM habit I have. can you explain why "via" is a better option?

sascala:
I can see why "via" could be problematic, since it's unnecessary latin. On the other hand, "by using" seemed vague. "Via" seemed to state more precisely what the user was doing with the GUI or CLI.

joel-hamill:
@sascala okay, thanks for the explanation.

PR #303, opened 2016-05-13
Documented third part containerizers support in roadmap.

haosdent:


haosdent:
Hi, @mhausenblas @guenter Is this make sense?

guenter:
@haosdent can you please file a JIRA ticket for this? The roadmap doc reflects what has been committed by the project. We need to prioritize this request first and we do that via JIRA. Hope that makes sense!

haosdent:
@guenter Got it, thanks a lot!

PR #304, opened 2016-05-13
removed empty cluster-size topic and links to it

joel-hamill:


emanic:
Oops, I cannot locate this branch anymore and notice it was located within the 1.7 subdirectory, which has gone away. Also, the files do not seem to exist anymore!

PR #305, opened 2016-05-13
Added enumeration topic

joel-hamill:


mhausenblas:
What's up with this one @sascala?

PR #307, opened 2016-05-13
fixup

joel-hamill:


emanic:
Sorry I am just now getting to this; I'm still figuring everything out. Thanks for preparing this edit. Actually, it's true that the user data is stored on Zookeeper, but not in this location. In discussing with Albert, we are thinking to leave out this information until we get some customer feedback where we see that people actually need it. It is only useful in troubleshooting-type situations, so not a normal use case. Unfortunately, this edit applies to a page that no longer exists. Closing this for now, but please let me know if you think we need this information.

joel-hamill:
@emanic I'm fine with leaving out for now.

emanic:
Oops, I am so sorry, I got this one mixed up with the Enterprise site. Sorry for the confusion!

emanic:
LGTM

PR #313, opened 2016-05-17
Fix Port numbers in docs

sargun:
Fix VIP without port number; For other VIP example, use different example port than 5050


mhausenblas:
@sargun can you resolve the conflicts pls?

PR #315, opened 2016-05-17
Suzanne/external storage

sascala:
Thought you might want to take a look at this before assigning it to Thomas. The External Volumes text is the same as on the Enterprise site, and the Local Volumes section was already there; the only thing to look over is index.md.

Also... I finally ran into the npm problem that you and Emily encountered. So I haven't been able to view the site on my local machine.


joel-hamill:
lgtm

PR #324, opened 2016-05-18
edits

joel-hamill:


joel-hamill:
@pyronicide so we do want to emphasize VIPs over Mesos DNS?

pyronicide:
@joel-hamill 100% yes

PR #326, opened 2016-05-18
fixed Mesos-DNS and Marathon capitalization

joel-hamill:


mhausenblas:
Nope.  @joel-hamill seems you did a search & replace everywhere and through that screwed up in a few places where it doesn't make sense or is downright wrong, for example it now reads `dcos Marathon app add`. Please roll back and do replace manually.

PR #334, opened 2016-05-20
Fix broken IP script

lloesche:
The script in `Use the network route to the Mesos master` is not doing what is described.

Instead of returning the interface IP it returns the gateway IP. Also it defines a variable `$MASTER_IP` but never uses it. Instead the IP is duplicated in the command. The `echo $()` would also make it so the script always exits with code 0 even when no valid IP could be found.

This PR fixes that. Also if no MASTER_IP is specified it uses the IP of which ever interface is connected to a default gateway.

```
[lukas@lukas.users ~]$ MASTER_IP=172.28.128.3
[lukas@lukas.users ~]$ echo $(/usr/sbin/ip route show to match 172.28.128.3 | grep -Eo '[0-9]{1,3}.[0-9]{1,3}.[0-9]{1,3}.[0-9]{1,3}' | tail -1)
192.168.200.1
[lukas@lukas.users ~]$ echo $(ip r g ${MASTER_IP:-8.8.8.8} | awk 'BEGIN {ec=1} {if($6=="src") {print $7; ec=0; exit}} END {exit ec}')
192.168.200.114
```

### DIFF
--- a/1.7/administration/installing/custom/advanced.md
+++ b/1.7/administration/installing/custom/advanced.md
@@ -107,11 +107,13 @@ The DC/OS installation creates these folders:
 
         ```bash
         #!/usr/bin/env bash
-        set -o nounset -o errexit
+        set -o nounset -o errexit -o pipefail
+        export PATH=/sbin:/usr/sbin:/bin:/usr/bin:$PATH
 
         MASTER_IP=172.28.128.3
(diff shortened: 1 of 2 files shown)


mhausenblas:
Thank you, @lloesche!

PR #339, opened 2016-05-24
Do not allow any filtering

sargun:


jgarcia-mesosphere:
LGTM (let's go to market) with this. I endorse these changes for CE.

sargun:
@mhausenblas

sargun:
@pyronicide

pyronicide:
This needs some love still. @sargun 

We need to list TCP, UDP and ICMP to make sure folks understand what "unfettered" actually means.

PR #347, opened 2016-05-26
moved topic to appropriate location

joel-hamill:


joel-hamill:
@pyronicide what do you think about this suggestion?

pyronicide:
@joel-hamill it doesn't make much sense to me. In the doc, we're showing how tasks are named and how you can figure that out. It doesn't have anything to do with nodes.

joel-hamill:
@pyronicide as I understand it, the purpose of this feature is so that users can figure out which node their installed services are running on. the current title "Discovering running service DNS entries" is very implementation-specific (and has back to back gerunds).

CC: @sargun

pyronicide:
No, the purpose of this feature is to figure out what DNS name your service has. Once you've figured out the DNS name, then you'd eventually use it to figure out the IP that your service has. That doesn't have anything to do with hosts though ... especially in an IP/container world.

joel-hamill:
@pyronicide how about `Discovering Service DNS Names`, or similar?

pyronicide:
That's getting closer. @sargun ?

1.7/administration/monitoring/view-node-details.md
```diff
@@ -0,0 +1,125 @@
+---
+post_title: Viewing node details
+nav_tile: Node details
+menu_order: 1
+---
+You can get a comprehensive list of the apps being run on your DC/OS cluster nodes.
+
+**Prerequisites:** [DC/OS and DC/OS CLI][1] are installed
+
+1.  SSH into your master node.
```
sargun: It can be any node.

1.7/administration/monitoring/view-node-details.md
```diff
@@ -0,0 +1,125 @@
...
+    
+        $ dcos node ssh --leader --master-proxy
+        
+    
+    For more information, the SSH [documentation][2].
+
+2.  Run this command from your master node to view the node details:
```
sargun: Should we suggest querying through AdminRouter? I wish we had a patched http or curl that added a token.

sargun:
Looks good.

1.7/administration/monitoring/view-node-details.md
joel-hamill: @sargun what are the instructions for doing this? why not just use this command, since it works?

joel-hamill:
@pyronicide ready to go

PR #353, opened 2016-05-31
moved crate to OSS

joel-hamill:
@pyronicide do we want this here? it no longer seems to fit on enterprise


pyronicide:
Have we tested that this works with 1.7 yet?

joel-hamill:
@pyronicide i have never tested either way. from my memory, this tutorial was written by vendor and they did testing/validation because we could not.

joel-hamill:
@keithchambers @kensipe any thoughts ^^

PR #354, opened 2016-05-31
corrected pub agent doc

joel-hamill:


emanic:
Hi Joel, I am a little bit uncertain as to what to do with this one...it is a file that is in a 1.7 subdirectory. It seems like if I merge this, "bad things may happen." Let me know the best way to proceed. Maybe I am missing something here. I don't have that directory anymore in my master. Thanks!

emanic:
The path was missing a slash. I have corrected this and pushed the change.

PR #357, opened 2016-06-01
Usage fixup

joel-hamill:


mhausenblas:
Why removing the 101 example? I don't think this is a good idea @joel-hamill

joel-hamill:
@mhausenblas a 101 example placed at the beginning of this top-level Usage topic is not good information design. a couple points: we cannot assume users have installed DC/OS (people don't read online documentation front-to-back); we cannot assume people are familiar with `docker run`.

it seems to make more sense to put this example directly at the end of the installation topics - since in those topics, we are sure that people have installed DC/OS and are ready to try it out. 

what do you think?

mhausenblas:
I don't know much about information design but I found it super useful. I leave it up to @pyronicide to decide …

mhausenblas:
Thanks @joel-hamill LGTM overall but two things we need to fix: 1. resolve conflicts, and 2. remove `- Running a [Ruby on Rails app](/docs/1.7/usage/tutorials/ruby-on-rails/)` since we pulled that tutorial.

PR #361, opened 2016-06-02
updated topic for completeness

joel-hamill:


joel-hamill:
merged to fast track for enterprise merge

PR #363, opened 2016-06-03
added missing menu order

joel-hamill:


joel-hamill:
needed to fast track for enterprise pull

PR #364, opened 2016-06-03
List fixup

joel-hamill:


joel-hamill:
fast track for enterprise merge

PR #365, opened 2016-06-03
Remove Ruby on Rails doc

jgarcia-mesosphere:
The Ruby on Rails doc depends on the MySQL doc, which has been removed. We should remove RoR as well.


mhausenblas:
It's still not applied since @pyronicide raised the issue that there are inbound links from other places in the docs that also need to be removed. Can I ask you to fix that @jgarcia-mesosphere and update the PR?

pyronicide:
I fixed the issues and pushed already. We need to just be more careful
before merging things like this in the future.

mhausenblas:
Ah, perfect, thank you @pyronicide and yes, totally agreed. Also: my bad.

PR #379, opened 2016-06-10
added full path to support sharing with enterprise

joel-hamill:
@karlkfi can we make an exception for this particular link?

### DIFF
--- a/1.7/administration/installing/local.md
+++ b/1.7/administration/installing/local.md
@@ -12,7 +12,7 @@ First, it's necessary to download the [DC/OS 1.7.0 Installer](https://downloads.
 
     $ curl -O https://downloads.dcos.io/dcos/EarlyAccess/dcos_generate_config.sh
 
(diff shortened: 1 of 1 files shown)


karlkfi:
Why? This will break dev.dcos.io

joel-hamill:
@karlkfi so we can reuse this topic here: https://docs.mesosphere.com/1.7/administration/installing/local/. alternatively we could remove from the enterprise doc (which might make sense).

karlkfi:
Long term, I'd like to see a /releases on mesosphere.com that provides links to the Enterprise downloads. Then the public change logs can go there and the download links can just be password protected.

joel-hamill:
@karlkfi what is dev.dcos.io used for? Staging?

karlkfi:
Yes. It's for verifying that changes to the website look correct before promoting them to production.

joel-hamill:
okay, i will stop sharing this topic, and update the link on enterprise.

PR #380, opened 2016-06-10
Collapses and cleans up services topics to facilitate synch with ente…

emanic:
…rprise site. Also adds back the cp option to CONTRIBUTING (corrected the command). Removes the BETA warning text from tutorial as it is no longer in beta."


joel-hamill:
@emanic did you search/replace for all links to this content in our doc as well? i suspect there could be quite a few. be sure to also create redirects in WP (for enterprise) and in this [file](https://github.com/dcos/dcos-website/blob/develop/redirects) (for OSS).

emanic:
@joel-hamill I only found two links, these were to the framework cleaner as discussed. Corrected these and also slight change to the cp command. I was not able to locate any other references...it really seems that people are only interested in linking to the framework cleaner content and they just repeat all the rest  of the content in their own service-specific topics rather than sending the reader to this central location. I think we might be good to go, let me know what you think.

joel-hamill:
@pyronicide we are planning to merge the 6 separate "Services" topic down to 1. There is currently a lot of overlap in these topics, and unnecessarily small topics. Thoughts?

pyronicide:
@joel-hamill I'm not a big fan, this feels like a step back in usability.

emanic:
This actually improves usability because the existing content has been edited to make many corrections, reduce redundancy, and generally increase clarity. In addition, it greatly increases the user experience in the enterprise site where the current Services is a mishmosh. This change will allow us to grow our existing service-specific content in the Enterprise site without sacrificing findability. It also institutes single-sourcing, which ensures that the content about all services is the same between the sites.

pyronicide:
That's all valid and doesn't address my point. By sticking all this into a single doc, it is hard for me to find and link to. We should definitely keep these identical between open/closed, the decision to put them into a single doc is not the right decision, imho.

emanic:
We have anchors allowing the links to go directly to the topic. Actually, many doc sites are moving to long-scrolling single page treatments because people can CTRL+F to find what they're looking for.

pyronicide:
I'd agree if we had the sidebar support. That's the biggest piece missing for me right now.

joel-hamill:
@pyronicide we do for all H1s, in enterprise WP doc :) http://cl.ly/0q0a2n3X2J24

can we chat about this, this afternoon?

mhausenblas:
@joel-hamill I suggest not to optimize so that it fits EE content best but that it serves our open source users first and foremost. The EE integration/transformation is a necessary step but that's the cost of doing business IMHO …

PR #407, opened 2016-06-22
how to print node details

joel-hamill:
What do you think about these shortened titles? We currently seem to overuse a lot of "DC/OS" instances.


1.7/usage/service-discovery/index.md
```diff
@@ -1,5 +1,5 @@
 ---
-post_title: Service Discovery and Load Balancing in DC/OS
```
pyronicide: The load balancing is important.

PR #410, opened 2016-06-23
Absolute nested path workaround for local persistent volumes

sascala:
@jieyu , if you have anything to add to the new section, please comment!

### DIFF
--- a/1.7/usage/storage/persistent-volume.md
+++ b/1.7/usage/storage/persistent-volume.md
@@ -34,7 +34,7 @@ Configure a persistent volume with the following options:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


joel-hamill:
@sascala can this be merged?

1.7/usage/storage/persistent-volume.md
```diff
@@ -34,24 +34,54 @@ Configure a persistent volume with the following options:
 }
 ```
 
-- `containerPath`: The path where your application will read and write data. This must be a single-level path relative to the container; it cannot contain a forward slash (`/`). (`"data"`, but not `"/data"`, `"/var/data"` or `"var/data"`).
+- `containerPath`: The path where your application will read and write data. This must be a single-level path relative to the container; it cannot contain a forward slash (`/`). (`"data"`, but not `"/data"`, `"/var/data"` or `"var/data"`). If your application requires absolute nested paths, [use this configuration](#nested-paths).
```
jieyu: If your application requires absolute or nested paths?

sascala:
@joel-hamill Jie had some comments I'd like to address before merging.

sascala:
@joel-hamill you can merge

PR #411, opened 2016-06-23
google analytics verification file

joel-hamill:


joel-hamill:
@pyronicide let me know when this change is live on the external site, and we begin using Google Analytics.

PR #416, opened 2016-06-30
Fixed marathon-lb menu order

sascala:
this should fix the menu order on the enterprise site.


sascala:
https://mesosphere.atlassian.net/browse/DOCS-947

PR #429, opened 2016-07-07
NFS for CoreOS

jgarcia-mesosphere:
Shared storage among nodes is a common use case. This guide walks users through a step-by-step procedure to configure the existing NFS components on the leader (NFS server) and agent (NFS client)


rji:
I think it's worth mentioning that this isn't a highly available setup.

joel-hamill:
@jgarcia-mesosphere is this applicable to Enterprise customers as well?

ssk2:
Applicable but, as Roger points, not HA. Most enterprise customers will have a storage appliance they use for NFS, so I'd recommend leaving this out of the enterprise docs.

PR #432, opened 2016-07-08
Adding instruction to set auth token

jibsonline:
Its a bit problematic as instructions to set the auth token is missing from the doc. It would be helpful if instructions are provided in the same tutorial


mhausenblas:
LGTM, thanks!

PR #438, opened 2016-07-13
Update docs on securing clusters.

brndnmtthws:
- renamed & relocated security doc
 - added instructions on redirecting HTTP to HTTPS with adminrouter
 - added note about open ports on public agents with CF templates

### DIFF
--- a/1.7/administration/securing-your-cluster.md
+++ b/1.7/administration/securing-your-cluster.md
@@ -1,6 +1,6 @@
 ---
-post_title: Security
-nav_title: Security
+post_title: Securing Your Cluster
+nav_title: Securing Your Cluster
 menu_order: 7
 ---
 
@@ -29,6 +29,44 @@ the other nodes in the cluster via URL routing. For security, the DC/OS cloud
 template allows configuring a whitelist so that only specific IP address
 ranges are permitted to access the admin zone.
 
+#### Steps for Securing Admin Router
+
+By default, Admin Router will permit unencrypted HTTP traffic. This is not
+considered secure, and you must provide a valid TLS certificate and redirect
+all HTTP traffic to HTTPS in order to properly secure access to your cluster.
+
+Once you have a valid TLS certificate, install the certificate on each master.
+Copy the certificate and private key to a well known location, such as under
+`/etc/ssl/certs`. It's suggested that you change the permissions of the
+private key to not be world-readable. Edit the Admin Router `nginx.conf`, by
+opening the file located at
+`/opt/mesosphere/active/adminrouter/nginx/conf`. Modify the nginx
+configuration `server` section, to look like this:
+
+```
+  server {
+    listen 80 default_server;
+    server_name _;
+    return 301 https://$host$request_uri;
+  }
+
+  server {
+      listen 443 ssl spdy default_server;
+      ssl_certificate /etc/ssl/certs/your-cert.crt;
+      ssl_certificate_key /etc/ssl/certs/your-key.key;
+
+      # rest of config intentionally omitted
+```
+
+In the block above, we're adding a new `server` section above the existing one
+which will redirect all HTTP traffic to HTTPS. We've also removed the
+`listen 80 ...` directive from the second server section. After making the
+changes, restart Admin Router:
+
+```console
+# systemctl restart dcos-adminrouter
+```
+
 ### Private zone
 
 The **private** zone is a non-routable network that is only accessible from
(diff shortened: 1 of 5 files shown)


1.7/administration/securing-your-cluster.md
```diff
@@ -29,6 +29,41 @@ the other nodes in the cluster via URL routing. For security, the DC/OS cloud
...
 
+#### Steps for Securing Admin Router
+
+By default, Admin Router will permit unencrypted HTTP traffic. This is not
+considered secure, and you must provide a valid SSL certificate and redirect
+all HTTP traffic to HTTPS in order to properly secure access to your cluster.
+
+Once you have a valid SSL certificate, install the certificate on each master.
+Edit the Admin Router `nginx.conf`, by opening the file located at
+`/opt/mesosphere/packages/adminrouter--*/nginx/conf`. Modify the nginx
```
adam-mesos: You can point them to `/opt/mesosphere/active/adminrouter/nginx/conf` so they don't have to worry about the package version sha

brndnmtthws: Nice, didn't know about this. Thanks1

1.7/administration/securing-your-cluster.md
```diff
@@ -47,6 +82,14 @@ only specific tasks can be scheduled here. These agent nodes have both public
 and private IP addresses and only specific ports should be open in their
 iptables firewall.
 
+By default, when using the cloud-based installers such as the AWS
+CloudFormation templates, a large number of ports are exposed to the Internet
+for the public zone. In production systems, it is unlikely that you would
+expose all of these ports. It's recommended that you close all ports except
+80 and 443 (for HTTP/HTTPS traffic) and use
```
adam-mesos: Why leave 80/HTTP open?

brndnmtthws: Even if you make everything HTTPS-only, you'll still have to redirect HTTP traffic to HTTPS (which requires port 80).

1.7/administration/securing-your-cluster.md
```diff
@@ -29,6 +29,41 @@ the other nodes in the cluster via URL routing. For security, the DC/OS cloud
 template allows configuring a whitelist so that only specific IP address
 ranges are permitted to access the admin zone.
 
+#### Steps for Securing Admin Router
+
+By default, Admin Router will permit unencrypted HTTP traffic. This is not
+considered secure, and you must provide a valid SSL certificate and redirect
+all HTTP traffic to HTTPS in order to properly secure access to your cluster.
+
+Once you have a valid SSL certificate, install the certificate on each master.
```
emanic: I wonder if it might be worth expanding a little bit here on how to install the cert on the masters? It seems that they can put it anywhere and just provide the path as noted below. What about the file permissions in that case and how should they protect the private key? Also, should the install the same cert on each master or different certs? These are some of the questions that come to my mind, at least.

brndnmtthws: I'll expand on this.

1.7/administration/securing-your-cluster.md
```diff
@@ -29,6 +29,44 @@ the other nodes in the cluster via URL routing. For security, the DC/OS cloud
 template allows configuring a whitelist so that only specific IP address
 ranges are permitted to access the admin zone.
 
+#### Steps for Securing Admin Router
+
+By default, Admin Router will permit unencrypted HTTP traffic. This is not
+considered secure, and you must provide a valid SSL certificate and redirect
+all HTTP traffic to HTTPS in order to properly secure access to your cluster.
```
emanic: I tend to like to use "TLS/SSL" as a qualifier instead of just "SSL." From what I understand, SSL is not really secure anymore and everyone should be using TLS. Though I am not yet sure which we are using or how to configure this. Just a minor comment as I think many people just use SSL when they really mean TLS or SSL.

adam-mesos: We're using TLS by default in Mesos/libprocess, although you can explicitly enable SSLv3 if you must.

brndnmtthws: Noted, I'll update accordingly.

1.7/administration/securing-your-cluster.md
```diff
@@ -1,6 +1,6 @@
 ---
-post_title: Security
-nav_title: Security
+post_title: Securing Your Cluster
+nav_title: Securing Your Cluster
```
emanic: I was just looking through the dcos.io site and I noticed the first paragraph of this page https://dcos.io/docs/1.7/administration/id-and-access-mgt/configure-your-security/ seems redundant with your new page. Although your new page is much more detailed on comprehensive. As such, it would seem to supercede this lonely paragraph. Perhaps we could delete that first paragraph then, if you agree, to avoid the redundancy. That way there's no confusion.

alberts:
lgtm. We'll make this automatic in 1.9, but good to document this process until then.

1.7/administration/securing-your-cluster.md
brndnmtthws: Makes sense, I'll drop the other paragraph.

emanic:
👍

PR #439, opened 2016-07-13
Moves usermgt page into id-and-access-mgt section; moves opt-out inst…

emanic:
…ructions for telemetry into telemetry topic; deletes opt-out.md as it is unnecessary and the auth section inside is redundant with opt-out instructions already in managing-auth topic; also adds sentence about upgrade option for opt-out in managing auth.


pyronicide:
I need a separate opt-out page and this is missing all the required redirects.

emanic:
Why do you need a separate opt-out page?

pyronicide:
To link users to, so that they know exactly how to opt-out without anything else on the page.

joel-hamill:
@pyronicide can you be more specific? from an information architecture perspective it's a bit odd to have Opt-Out on the same heading level/importance as Logging or Installation.

pyronicide:
Some users care about everything they can turn off (and want to). Note that this refactor looses information (it doesn't tell how to opt-out of auth and doesn't include the redirects to make sure all the product links continue to work).

emanic:
Just to clarify, it does not lose any information. Actually, the content was repeated in https://dcos.io/docs/1.7/administration/id-and-access-mgt/managing-authentication/ So, this actually removes a redundancy.

joel-hamill:
Closing. These PRs duplicate effort: https://github.com/dcos/dcos-docs/pull/479 and https://github.com/dcos/dcos-docs/pull/560

PR #444, opened 2016-07-18
added new overlay network graphic

sascala:
I updated the graphic with the lovely new one created by design. NB that I can't preview from my local machine because `1.8/` isn't "latest".


emanic:
Wow, nice. 👍  Too bad no preview yet, but can always preview on live site, I guess.

PR #447, opened 2016-07-20
removed line breaks/spaces

joel-hamill:


joel-hamill:
hard line breaks don't behave well on the enterprise docs: https://cl.ly/1k0N0s0p3Q1E.

PR #454, opened 2016-07-22
Fix invalid front matter

TimonVS:


1.7/administration/storage/mount-disk-resources.md
```diff
@@ -1,7 +1,7 @@
 ---
 post_title: Mount Disk Resources
 menu_order: 0
--------------
```
karlkfi: why does this fail parsing?

TimonVS: I think it interprets the extra dashes as a list or something. It really depends on which parser is being used.

PR #456, opened 2016-07-22
pulling 1.8 branch off of master

joel-hamill:


pyronicide:
Why?

joel-hamill:
@pyronicide because we have not EA/GA'd yet. although is there a magic build that doesn't use everything in master?

PR #458, opened 2016-07-25
Use lowercase only for HAPROXY_0_VHOST label value

surdy:
I followed the doc and was not able to make the VHOST setting work until I changed the label value to all lowercase. Updated the marathon JSON to reflect it. 

CC @brndnmtthws for review.


brndnmtthws:
Strange, it shouldn't matter since the checks are case insensitive. Regardless, 👍 and thanks!

PR #462, opened 2016-07-26
Remove redundant statement from M-LB advanced doc

surdy:
First two sentences on Point `1.` are redundant (and confusing)


joel-hamill:
thanks!

PR #478, opened 2016-08-01
Update ip-per-container.md

fernandosanchezmunoz:
add clarification about why the 6 bits used for containerID are divided into two groups of 5 bits, one for Mesos and the other for Docker containers


sascala:
@fernandosanchezmunoz I incorporated your edit here: https://github.com/dcos/dcos-docs/pull/569

PR #480, opened 2016-08-02
Limitations to number of containers on overlay network

sascala:
Thanks for your help, @asridharan I'm going to ask to merge this without the example we discussed. Let's add an example for 1.8 GA.

https://mesosphere.atlassian.net/browse/DOCS-1039


asridharan:
👍

PR #486, opened 2016-08-03
Ss/ui fixes

sascala:
Despite some outstanding questions about VIPs, I think this is ready for review.


sascala:
Thanks for your review, @leemunroe !

PR #494, opened 2016-08-04
Added link to list of time zones to jobs doc

sascala:
@pierlo-upitup Is this what you were thinking?

### DIFF
--- a/1.8/usage/jobs/getting-started.md
+++ b/1.8/usage/jobs/getting-started.md
@@ -16,7 +16,7 @@ From the DC/OS web interface, click the **Jobs** tab, then the **New Job** butto
 * **Memory**
 * **Disk space**
 * **Command** - The command your job will execute. Leave this blank if you will use a Docker image.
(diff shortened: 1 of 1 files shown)


pierlo-upitup:
Thanks @sascala – how about adding one line to explain the cron format supported by metronome? Afaik there are some [variations](https://en.wikipedia.org/wiki/Cron#CRON_expression), and as a user I'd find great value in being told which format is supported.

@aquamatthias which library is metronome using to parse these strings? Maybe we can take it from there.

sascala:
@aquamatthias Do you happen to know which cron format Metronome supports?

aquamatthias:
@sascala We support the standard 5 element cron expression:
(minute hour dayOfMonth month dayOfWeek)

We also support an optional year at the end:
(minute hour dayOfMonth month dayOfWeek, year)

See https://github.com/dcos/metronome/blob/master/jobs/src/main/scala/dcos/metronome/model/CronSpec.scala#L32

emanic:
LGTM

PR #502, opened 2016-08-05
The script link not working

jibsonline:
The link for the script is not working
the script execution will throw error due to permission issues. so it has to be executed this way


joel-hamill:
@ashenden i think this is caused by the same issue we had with `.json` files

ashenden:
Yup.  Seems like the build script compiles `.md` files into `/directory/index.html`.  All other file types have to be manually defined in the build script by their file type, otherwise they aren't moved into the build folder.  So anything with `.sh` we want copied to the build folder?

pyronicide:
Yup, anything with `.sh`.

joel-hamill:
@ashenden ^

joel-hamill:
the underlying problems seems to have been fixed. closing without merging.

PR #511, opened 2016-08-09
use --remote-name when downloading dcos binary.

orsenthil:
I tried the `curl` command suggested in the docs and the binary was streamed to my console. Better to suggest `curl --remote-name` to the customer so that they can download the binary.


joel-hamill:
@tamarrow what do you think?

tamarrow:
The previous UI instructions use `-O`, lets keep that to be consistent. We should also add this to the mac section

orsenthil:
So, the change suggested is to use `-O` instead of `--remote-name` ?  I will update.

joel-hamill:
thanks!

PR #512, opened 2016-08-09
Fixed networking organization

sascala:
Moved overlay networks to svc disc > load balancing + vips; deleted name-based vips


joel-hamill:
@sascala does the link to overlay-networks need to be updated too: https://github.com/dcos/dcos-docs/blob/7fd3c1a21e646c3d34f5233fcf6f1356b545fcd0/1.8/administration/overlay-networks/index.md

sascala:
@joel-hamill No, that section is called "Overlay Networks" for the moment, since we don't have any other VIP functionality at the moment. For usage, the topic fits into service discovery and load balancing, but the administration section doesn't have that.

PR #546, opened 2016-08-24
add release notes to main doc TOC

joel-hamill:


cruhland:
@joel-hamill I see that the file is located in the `1.8/administration/release-notes` directory. Based on the layout of the files under `1.8`, it looks like directories translate into "menu groups" on the website. Shouldn't the release notes link just be a single document underneath `1.8/administration`, instead of having its own subdirectory/menu-group?

cruhland:
:ship: it!

PR #553, opened 2016-08-28
DCOS-343 - Document Exhibitor backend for Azure

justenwalker:
This PR addresses DCOS-343 by adding adding the `azure` backend as another exhibitor storage backend, and documenting it's required properties.


justenwalker:
I think my editor may have removed some whitespace from empty lines. They don't appear to be necessary; however, if it's a problem, I can undo those changes.

lingmann:
LGTM thanks for the PR @justenwalker !

PR #571, opened 2016-09-07
Change CLI install instructions

jsancio:
Changes:
1. Changed the URL for the binary to be pinged to the dcos release version.
2. Remove `$` from the examples so that the user can copy and paste them.
3. Minor removal of empty white spaces.


jsancio:
cc @tamarrow

jsancio:
I ran through the instructions in Linux and they worked.

PR #580, opened 2016-09-12
corrected upgrade instructions

joel-hamill:


1.8/usage/cli/update.md
```diff
@@ -12,17 +12,56 @@ You can update the DC/OS CLI to the latest version or downgrade to an older vers
 
 You can upgrade an existing DC/OS CLI installation to the latest build.
 
-1.  Remove the current CLI binary:
-
+1.  Remove the current CLI binary. By default, this file is located in `/usr/local/bin/dcos`.
```
tamarrow: we recommend instead of default? default is really 'they copied the UI instructions blindly`

1.8/usage/cli/update.md
```diff
@@ -12,17 +12,56 @@ You can update the DC/OS CLI to the latest version or downgrade to an older vers
...
+                                   
     ```bash
-    $ rm path/to/binary/dcos
+    $ rm -rf /usr/local/bin/dcos
     ```
 
-1.  From the directory you want to install the new DC/OS CLI binary, enter this command to update the DC/OS CLI with the upgrade version (`<version>`) specified:
+1.  Download the DC/OS CLI binary (`dcos`) to your local directory (for example, `/usr/local/bin/`).Update the command with the desired upgrade version (`<version>`):
```
tamarrow: space after period

1.8/usage/cli/update.md
```diff
@@ -12,17 +12,56 @@ You can update the DC/OS CLI to the latest version or downgrade to an older vers
...
     
     ```bash
     $ curl https://downloads.dcos.io/binaries/cli/darwin/x86-64/<version>/dcos
     ```
+    
+    **Important:** The CLI must be installed on a system that is external to your DC/OS cluster.
+    
+1.  Make the CLI binary executable.
```
tamarrow: (on mac and linux only)

1.8/usage/cli/update.md
```diff
@@ -12,17 +12,56 @@ You can update the DC/OS CLI to the latest version or downgrade to an older vers
...
+    
+    ```bash
+    chmod +x dcos
+    ```
+    
+1.  Point the CLI to your DC/OS master node. In this example, `http://example.com` is the master node IP address.
+    
+    ```bash
+    ./dcos config set core.dcos_url http://example.com
```
tamarrow: if the directory the cli is installed is already in `$PATH` they don't need the `./` before dcos. Maybe step before would be to add the directory to PATH if not already

1.8/usage/cli/update.md
joel-hamill: how do Windows users do this?

tamarrow: windows binaries have the .exe extension so they are already executable

joel-hamill:
@tamarrow i have updated with your comments. thanks.

PR #583, opened 2016-09-13
Update upgrade doc for 1.8

branden:


1.8/administration/upgrading.md
```diff
@@ -70,7 +70,12 @@ Identify your Mesos leader node. This node should be the last master node that y
     $ sudo rm -rf /opt/mesosphere /etc/mesosphere
     ```
 
-<!-- Insert additional step for `exhibitor_storage_backend: zookeeper` users. -->
+1. **Important:** If `exhibitor_storage_backend` is set to `zookeeper`, add the user `dcos_exhibitor` and give it ownership of the ZooKeeper data directory:
+
+    ```
+    $ sudo useradd --system --home-dir /opt/mesosphere --shell /sbin/nologin -c 'DCOS System User' dcos_exhibitor
```
branden: @cmaloney LMK if this `useradd` isn't right.

cmaloney: https://github.com/dcos/dcos/blob/master/pkgpanda/__init__.py#L540 -- should do exactly what that does, which seems to be what this is doing.

cmaloney:
Fixes: https://dcosjira.atlassian.net/browse/DCOS-306

1.8/administration/upgrading.md
```diff
@@ -70,7 +70,12 @@ Identify your Mesos leader node. This node should be the last master node that y
...
+    $ sudo chown -R dcos_exhibitor /var/lib/zookeeper
+    ```
```
cmaloney: Need to add steps here to after upgrading, go into the exhibitor UI, and update the zookeeper config to point to the new zk location (`/var/lib/dcos/exhibitor/zookeeper` I think). If you launch a fresh cluster that has the new config.

Otherwise this will bite even harder on future upgrades.

branden: https://mesosphere.atlassian.net/browse/DCOS-9782

PR #596, opened 2016-09-15
Ss/octarine

sascala:
@nlsun Could you please read this over for accuracy? Also, if there is any other information a user might require for success, please let me know!

### DIFF
--- a/1.8/administration/access-node/tunnel.md
+++ b/1.8/administration/access-node/tunnel.md
@@ -1,6 +1,6 @@
 ---
-nav_title: Access by Proxy
-post_title: Access by Proxy using DC/OS Tunnel
+nav_title: Access by Proxy and VPN
+post_title: Access by Proxy and VPN using DC/OS Tunnel
 menu_order: 10
 ---
 
--- a/1.8/administration/access-node/tunnel.md
+++ b/1.8/administration/access-node/tunnel.md
@@ -61,7 +61,7 @@ DC/OS Tunnel provides you with full access to the DNS, masters, and agents from 
       </td>
       <td>
           <ul>
-              <li>Cannot specify port (except through SRV)</li>
+              <li>Cannot specify ports (except through SRV)</li>
               <li>Only supports HTTP</li>
               <li>Runs as superuser</li>
           </ul>
--- a/1.8/administration/access-node/tunnel.md
+++ b/1.8/administration/access-node/tunnel.md
@@ -79,6 +79,7 @@ DC/OS Tunnel provides you with full access to the DNS, masters, and agents from 
           <td>
               <ul>
                   <li>Requires application configuration</li>
+                  <li>Only supports HTTP/HTTPS</li>
               </ul>
               </td>
      </tr>
--- a/1.8/administration/access-node/tunnel.md
+++ b/1.8/administration/access-node/tunnel.md
@@ -28,7 +28,7 @@ DC/OS Tunnel provides you with full access to the DNS, masters, and agents from 
 * For VPN functionality, you must have [the OpenVPN client](https://openvpn.net/index.php/open-source/downloads.html) installed.
 * The DC/OS Tunnel package must be installed: run `dcos package install tunnel-cli --cli` from the DC/OS CLI.
 
-**Note:** * Only Linux and OS X are supported currently.
+**Note:** Only Linux and OS X are supported currently.
 
 ##  SOCKS
 Run the following command from the DC/OS CLI:
@@ -40,7 +40,7 @@ $ dcos tunnel socks
 Configure your application to use the proxy on port 1080: `127.0.0.1:1080`
 
 ### Port Forwarding
-The SOCKS proxy works by port forwarding. Append `.mydcos.directory` to the end of your domain when you enter commands. For instance, `http://example.com:8080/?query=hello` becomes `http://example.com.mydcos.directory:1080/?query=hello`.
+The SOCKS proxy works by port forwarding. Append `.mydcos.directory` to the end of your domain when you enter commands. For instance, `http://example.com:1080/?query=hello` becomes `http://example.com.mydcos.directory:1080/?query=hello`.
 
 ##  HTTP
 Run the following command from the DC/OS CLI:
@@ -49,7 +49,7 @@ Run the following command from the DC/OS CLI:
 $ sudo dcos tunnel http
 ```
 
-To run the HTTP proxy without root privileges, configure it not to use port 80:
+To run the HTTP proxy without root privileges, use the `--port` flag to configure it to use another port:
 
 ```
 $ dcos tunnel http --port 8000
@@ -63,23 +63,33 @@ The HTTP proxy works by port forwarding. Append `.mydcos.directory` to the end o
 ### SRV Records
 The HTTP proxy exposes DC/OS SRV records as URLs in the form `_<port-name>._<service-name>._tcp.marathon.mesos.mydcos.directory`.
 
-To name a port, add `name` to the `portMappings` field of a Marathon application definition:
+To name a port, add `name` to the `portMappings` or `portDefinitions` field of a Marathon application definition. Whether you use `portMappings` or `portDefinitions` depends on whether you are using `BRIDGE` or `HOST` networking. [Learn more about networking and ports in Marathon](https://mesosphere.github.io/marathon/docs/ports.html).
 
 ```json
 "portMappings": [
-        {
-          "name": "<my-port-name>",
-          "containerPort": 3000,
-          "hostPort": 0,
-          "servicePort": 10000,
-          "labels": {
-            "VIP_0": "1.1.1.1:30000"
-          }
+    {
+        "name": "<my-port-name>",
+        "containerPort": 3000,
+        "hostPort": 0,
+        "servicePort": 10000,
+        "labels": {
+             "VIP_0": "1.1.1.1:30000"
         }
-      ]
+    }
+]
 ```
 
-Alternatively, add a named port from the DC/OS web interface by going to **Services** > **Deploy Service**. Then, click the **Network** tab to enter a name for your port.
+```json
+"portDefinitions": [
+    {
+      "name": "<my-port-name>",
+      "protocol": "tcp",
+      "port": 0,    
+    }
+  ]
+```
+
+Alternatively, you can add a named port from the DC/OS web interface. Go to the **Services** tab, click the name of your service, and then click **Edit**. Enter a name for your port on the **Network** tab.
 
 The `<service-name>` is the value of the `id` field in your Marathon application definition or the entry in the **ID** field of a service you create from the DC/OS web interface.
 
@@ -97,7 +107,7 @@ The VPN does not work by port forwarding, so you do not need to append `.mydcos.
 When you use the VPN, you are virtually within your cluster. To access your master node, just enter the following from your terminal:
 
 ```
-ssh core@master.mesos
+$ ssh core@master.mesos
 ```
 
 Similarly, you can interact with your agent nodes from virtually within the cluster. For instance:
--- a/1.8/administration/access-node/tunnel.md
+++ b/1.8/administration/access-node/tunnel.md
@@ -39,9 +39,6 @@ $ dcos tunnel socks
 
 Configure your application to use the proxy on port 1080: `127.0.0.1:1080`
 
-### Port Forwarding
-The SOCKS proxy works by port forwarding. Append `.mydcos.directory` to the end of your domain when you enter commands. For instance, `http://example.com:1080/?query=hello` becomes `http://example.com.mydcos.directory:1080/?query=hello`.
-
 ##  HTTP
 Run the following command from the DC/OS CLI:
 
@@ -61,9 +58,16 @@ Then, configure your application to run HTTP on the port you specified above.
 The HTTP proxy works by port forwarding. Append `.mydcos.directory` to the end of your domain when you enter commands. For instance, `http://example.com:8080/?query=hello` becomes `http://example.com.mydcos.directory:8080/?query=hello`.
 
 ### SRV Records
-The HTTP proxy exposes DC/OS SRV records as URLs in the form `_<port-name>._<service-name>._tcp.marathon.mesos.mydcos.directory`.
+The HTTP proxy exposes DC/OS SRV records as URLs in the form `_<port-name>._<service-name>._tcp.marathon.mesos`.
 
-To name a port, add `name` to the `portMappings` or `portDefinitions` field of a Marathon application definition. Whether you use `portMappings` or `portDefinitions` depends on whether you are using `BRIDGE` or `HOST` networking. [Learn more about networking and ports in Marathon](https://mesosphere.github.io/marathon/docs/ports.html).
+#### Find your Service Name
+The `<service-name>` is the entry in the **ID** field of a service you create from the DC/OS web interface or the value of the `id` field in your Marathon application definition.
+
+#### Add a Named Port from the DC/OS Web Interface
+To name a port from the DC/OS web interface, go to the **Services** tab, click the name of your service, and then click **Edit**. Enter a name for your port on the **Network** tab.
+
+#### Add a Named Port in a Marathon Application Definition
+Alternatively, you can add `name` to the `portMappings` or `portDefinitions` field of a Marathon application definition. Whether you use `portMappings` or `portDefinitions` depends on whether you are using `BRIDGE` or `HOST` networking. [Learn more about networking and ports in Marathon](https://mesosphere.github.io/marathon/docs/ports.html).
 
 ```json
 "portMappings": [
@@ -89,10 +93,6 @@ To name a port, add `name` to the `portMappings` or `portDefinitions` field of a
   ]
 ```
 
-Alternatively, you can add a named port from the DC/OS web interface. Go to the **Services** tab, click the name of your service, and then click **Edit**. Enter a name for your port on the **Network** tab.
-
-The `<service-name>` is the value of the `id` field in your Marathon application definition or the entry in the **ID** field of a service you create from the DC/OS web interface.
-
 ## VPN
 Run the following command from the DC/OS CLI
 
@@ -101,8 +101,6 @@ $ sudo dcos tunnel vpn
 ```
 
 The VPN client attempts to auto-configure DNS, but this functionality does not work on Mac OSX. To use the VPN client on OSX, [add the DNS servers](https://support.apple.com/kb/PH18499?locale=en_US) that DC/OS Tunnel instructs you to use.
-
-The VPN does not work by port forwarding, so you do not need to append `.mydcos.directory` to the end of your commands. For example, to access the DC/OS web interface from within the cluster, point your browser to `https://master.mesos`.
 
 When you use the VPN, you are virtually within your cluster. To access your master node, just enter the following from your terminal:
 
--- a/1.8/administration/access-node/tunnel.md
+++ b/1.8/administration/access-node/tunnel.md
@@ -4,20 +4,107 @@ post_title: Access by Proxy and VPN using DC/OS Tunnel
 menu_order: 10
 ---
 
-When developing applications on DC/OS, you may find it helpful to access your cluster from your local machine via SOCKS proxy, HTTP proxy, or VPN. For instance, you can work from your own development environment and immediately test against your DC/OS cluster.
+When developing services on DC/OS, you may find it helpful to access your cluster from your local machine via SOCKS proxy, HTTP proxy, or VPN. For instance, you can work from your own development environment and immediately test against your DC/OS cluster.
 
 # SOCKS
-DC/OS Tunnel can run a SOCKS proxy over SSH to the cluster. SOCKS proxies work for any protocol, but your application must be configured to use the proxy, which runs on port 1080 by default.
+DC/OS Tunnel can run a SOCKS proxy over SSH to the cluster. SOCKS proxies work for any protocol, but your service must be configured to use the proxy, which runs on port 1080 by default.
 
 # HTTP
-Run the HTTP proxy on your local port 80 to tunnel HTTP requests without modifying your application. Access URLs by appending the `mydcos.directory` domain. You can also [use DNS SRV records as if they were URLs](#srv). The HTTP proxy requires root access to use port 80, but can be configured to use a different port.
+
+The HTTP proxy can run in two modes: transparent and standard.
+
+##  Transparent Mode
+In transparent mode, the HTTP proxy runs as superuser on port 80 and does not require modification to your application. Access URLs by appending the `mydcos.directory` domain. You can also [use DNS SRV records as if they were URLs](#srv).
+
+## Standard Mode
+You must configure your service to use the HTTP proxy in standard mode, though it does not have any of the limitations of transparent mode. As in transparent mode, you can use [DNS SRV](#srv) records as URLs.
 
 <a name="srv"></a>
 ## SRV Records
 A SRV DNS record is a mapping from a name to a IP/port pair. DC/OS creates SRV records in the form `_<port-name>._<service-name>._tcp.marathon.mesos`. The HTTP proxy exposes these as URLs. This feature can be useful for communicating with DC/OS services.
 
+**Note:** The HTTP proxy cannot currently access HTTPS.
+
 # VPN
 DC/OS Tunnel provides you with full access to the DNS, masters, and agents from within the cluster. OpenVPN requires root privileges to configure these routes.
+
+# DC/OS Tunnel Options at a Glance
+
+<table class="table">
+  <tr>
+    <th>&nbsp;</th>
+    <th>Pros</th>
+    <th>Cons</th>
+  </tr>
+  <tr>
+    <th>SOCKS</th>
+    <td>
+    <ul>
+        <li>Specify ports</li>
+        <li>All protocols</li>
+    </ul>
+    </td>
+    <td>
+        <ul>
+            <li>Requires application configuration</li>
+        </ul>
+        </td>
+  </tr>
+  
+  <tr>
+      <th>HTTP (transparent)</th>
+      <td>
+      <ul>
+          <li>SRV as URL</li>
+          <li>No application configuration</li>
+      </ul>
+      </td>
+      <td>
+          <ul>
+              <li>Cannot specify port</li>
+              <li>Only supports HTTP</li>
+              <li>Runs as superuser</li>
+          </ul>
+          </td>
+    </tr>
+    
+    <tr>
+          <th>HTTP (standard)</th>
+          <td>
+          <ul>
+              <li>SRV as URL</li>
+              <li>Specify ports</li>
+          </ul>
+          </td>
+          <td>
+              <ul>
+                  <li>Requires application configuration</li>
+                  <li>Only supports HTTP</li>
+              </ul>
+              </td>
+     </tr>
+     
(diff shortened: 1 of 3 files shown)


1.8/administration/access-node/tunnel.md
```diff
@@ -1,50 +1,113 @@
...
+Run the following command from the DC/OS CLI:
+
+```
+$ dcos tunnel socks
+```
+
+Configure your application to use the proxy on port 1080: `127.0.0.1: 1080`
+
+### Port Forwarding
+The SOCKS proxy works by port forwarding. Append `.mydcos.directory` to the end of your domain when you enter commands. For instance, `http://example.com:8080/?query=hello` becomes `http://example.com.mydcos.directory:8080/?query=hello`.
```
sascala: Should this be `http://example.com:1080`?

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,113 @@
+---
+post_title: Access by Proxy
+menu_order: 10
+---
+
+When developing applications on DC/OS, you may find it helpful to access your cluster via SOCKS proxy, HTTP proxy, or VPN. For instance, you can work from your own development environment and immediately test on your DC/OS cluster.
```
nlsun: I think this might make more sense as "test against" rather than "test on". "test on" makes it sound like the proxy will help you run stuff on DC/OS rather than "testing against" a service that is already running on DC/OS (using local debugging tools from your development environment)

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,113 @@
...
+
+# SOCKS
+dcos tunnel can run a SOCKS proxy over SSH to the cluster. SOCKS proxies work for any kind of network protocol on any port, but your application must be configured to use the proxy, which runs on port 1080 by default. Access URLs by appending the `mydcos.directory` domain.
```
nlsun: Instead of "SOCKS proxies work for any kind of network protocol on any port" I'd just say "SOCKS proxies work for any protocol". It doesn't deal with the network protocols, a more pedantic version would be "application protocol" I suppose, but I think just plain "protocol" is more easily understood.

And all of the proxies can access all ports so that isn't particular to this one.

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,113 @@
...
+
+# HTTP
+Run the HTTP proxy on your local port 80 to tunnel HTTP requests without modifying your application. Access URLs by appending the `mydcos.directory` domain. You can also [use DNS SRV records as if they were URLs](#srv).
+
+<a name="srv"></a>
+## SRV Records
+An SRV DNS record is a mapping from a name to a IP/port pair. DC/OS creates SRV records in the form `_<port-name>._<task-name>._tcp.marathon.mesos.mydcos.directory`. The HTTP proxy exposes these as URLs. This feature can be useful for communicating with Marathon apps.
```
nlsun: "An" -> "A"

nlsun: The SRV records are actually `_<port-name>._<task-name>._tcp.marathon.mesos`. The `.mydcos.directory` is just due to the URL access syntax. (Sorry, just realized I made this same mistake in the doc I gave you).

nlsun: Maybe instead of "Marathon apps" we should call them "DC/OS Services" (I think that's what they're called, may want to verify that with someone else as well).

nlsun: Maybe instead of "Marathon apps" we should call them "DC/OS Services" (I think that's what they're called, may want to verify that with someone else as well).

1.8/administration/access-node/tunnel.md
nlsun: It was referred to several times as "dcos tunnel" but maybe they should all be changed to "DC/OS Tunnel" or something more official sounding?

1.8/administration/access-node/tunnel.md
sascala: Agreed. And yes, AFAIK that's what they're called.

1.8/administration/access-node/tunnel.md
sascala: I'd been unsure what to call it. I'll confer with @joel-hamill

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,113 @@
...
+menu_order: 10
+---
+
+When developing applications on DC/OS, you may find it helpful to access your cluster via SOCKS proxy, HTTP proxy, or VPN. For instance, you can work from your own development environment and immediately test on your DC/OS cluster.
+
+# SOCKS
+dcos tunnel can run a SOCKS proxy over SSH to the cluster. SOCKS proxies work for any kind of network protocol on any port, but your application must be configured to use the proxy, which runs on port 1080 by default. Access URLs by appending the `mydcos.directory` domain.
+
+# HTTP
+Run the HTTP proxy on your local port 80 to tunnel HTTP requests without modifying your application. Access URLs by appending the `mydcos.directory` domain. You can also [use DNS SRV records as if they were URLs](#srv).
```
nlsun: Since you mention that the VPN requires root access, you might also want to mention that using the port 80 access here will also require root access. However, you can also change the port (via the `--port` flag) to one that doesn't require root access.

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,113 @@
...
+##  SOCKS
+Run the following command from the DC/OS CLI:
+
+```
+$ dcos tunnel socks
+```
+
+Configure your application to use the proxy on port 1080: `127.0.0.1:1080`
+
+### Port Forwarding
```
nlsun: May be better to have this block once, and link to it from the SOCKS and HTTP topics

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,113 @@
...
+The HTTP proxy works by port forwarding. Append `.mydcos.directory` to the end of your domain when you enter commands. For instance, `http://example.com:8080/?query=hello` becomes `http://example.com.mydcos.directory:8080/?query=hello`.
+
+### SRV Records
+The HTTP proxy exposes DC/OS SRV records as URLs in the form `_<port-name>._<task-name>._tcp.marathon.mesos.mydcos.directory`.
+
+To name a port, add `name` to the `portMappings` field of a Marathon application definition:
+
+```json
+"portMappings": [
```
nlsun: Hmm it kinda looks like `containerPort`, `hostPort`, etc are related to naming the port. Maybe just mention that `name` can be added to both `portMappings` and `portDefinitions`?

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,113 @@
...
+          "servicePort": 10000,
+          "name": "<my-port-name>",
+          "labels": {
+            "VIP_0": "1.1.1.1:30000"
+          }
+        }
+      ]
+```
+
+Alternatively, [add a named VIP to your application from the DC/OS web interface](https://dcos.io/docs/1.8/usage/service-discovery/load-balancing-vips/virtual-ip-addresses/).
```
nlsun: SRV records aren't related to named VIPS

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,113 @@
...
+The `<task-name>` is the value of the `id` field in your Marathon application definition or the entry in the **ID** field of a service you create from the DC/OS web interface.
+
+## VPN
+Run the following command from the DC/OS CLI
+
+```
+$ sudo dcos tunnel vpn
+```
+
+The VPN client auto-configures DNS, but this functionality does not work on Mac OSX. To use the VPN client on OSX, [add the following DNS servers](https://support.apple.com/kb/PH18499?locale=en_US):
```
nlsun: maybe say "The VPN client attempts to auto-configure DNS"

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,113 @@
...
+
+```
```
nlsun: Oh, this one shouldn't be hard-coded here, the CLI should give them instructions appropriate to their cluster.

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,113 @@
...
+198.51.100.1
+198.51.100.2
+198.51.100.3
+```
+
+The VPN does not work by port forwarding, so you do not need to append `.mydcos.directory` to the end of your commands. For example, to access the DC/OS web interface from within the cluster, point your browser to `https://marathon.mesos`.
```
nlsun: Maybe we want to have `master.mesos` instead of `marathon.mesos` here

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,113 @@
...
+
+When you use the VPN, you are virtually within your cluster. To access your master node, just enter the following from your terminal:
+
+```
+ssh core@master.mesos
+```
+
+Similarly, you can interact with your agent nodes from virtually wthin the cluster. For instance:
```
nlsun: "wthin" > "within"

1.8/administration/access-node/tunnel.md
nlsun: Ah, didn't mean for this to look similar to anything. Maybe using 5555 or something will make it more clear that this is just a dummy example.

nlsun: Oh, I realize now that this paragraph was specific to SOCKS, I think port forwarding can be a separate thing that gets reused between HTTP and SOCKS.

1.8/administration/access-node/tunnel.md
sascala: 👍

1.8/administration/access-node/tunnel.md
sascala: DC/OS Tunnel it is.

1.8/administration/access-node/tunnel.md
sascala: I get into that in the usage section. Added a mention of that here too.

1.8/administration/access-node/tunnel.md
sascala: @karlkfi made the point that `<task-name>` is a little misleading here and might be more appropriately called `<service-name>`. Any objection?

1.8/administration/access-node/tunnel.md
sascala: I prefer to repeat because users don't always read the whole doc and clicking around seems cumbersome.

1.8/administration/access-node/tunnel.md
sascala: Agreed that maybe supplying the complete portMappings JSON could be misleading.

Can you explain more about `portDefintions`? Can a user set both in their app definition? Only one? When would they choose `portDefinitions` over `portMappings`?

1.8/administration/access-node/tunnel.md
nlsun: sounds good to me

1.8/administration/access-node/tunnel.md
sascala: My bad.
Is there a way to set the port name from the DC/OS UI?

1.8/administration/access-node/tunnel.md
nlsun: AFAIK you can only set one. But I don't think this is the place to explain those details. Is there another doc page that mentions how to name a port?

1.8/administration/access-node/tunnel.md
sascala: I don't have a cluster up anymore -- does that also go to the DC/OS UI?

1.8/administration/access-node/tunnel.md
nlsun: Yep, in the create new service modal, in the "Network" tab there's a field for a "Name"

1.8/administration/access-node/tunnel.md
nlsun: Yup

sascala:
@nlsun I think I've taken care of your comments. Thanks a lot for your careful reading.

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,118 @@
...
+Run the following command from the DC/OS CLI:
+
+```
+$ dcos tunnel socks
+```
+
+Configure your application to use the proxy on port 1080: `127.0.0.1:1080`
+
+### Port Forwarding
+The SOCKS proxy works by port forwarding. Append `.mydcos.directory` to the end of your domain when you enter commands. For instance, `http://example.com:1080/?query=hello` becomes `http://example.com.mydcos.directory:1080/?query=hello`.
```
nlsun: Oops, I just realized, the port you access in the URL is unrelated to the port that the SOCKS proxy is running on. The SOCKS proxy port is local (e.g. on your laptop), and the URL port is remote (e.g. on your DC/OS cluster). So this port shouldn't be the SOCKS port to avoid being misleading. Sorry for flipping back and forth on this point.

The same goes for the Port Forwarding under the HTTP part. They should both be some arbitrary port

nlsun: Ahhh. I just realized that the SOCKS proxy doesn't need the port forwarding at all. It's just that it _also_ works with port forwarding. In any case, please remove anything that relates `mydcos.directory` to SOCKS since it just complicates things.

Only the HTTP proxy needs `mydcos.directory`. Sorry.

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,118 @@
...
+"portDefinitions": [
+    {
+      "name": "<my-port-name>",
+      "protocol": "tcp",
+      "port": 0,    
+    }
+  ]
+```
+
+Alternatively, you can add a named port from the DC/OS web interface. Go to the **Services** tab, click the name of your service, and then click **Edit**. Enter a name for your port on the **Network** tab.
```
nlsun: Maybe we should make the UI way the primary way and the json way the alternative way?

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,118 @@
...
+## VPN
+Run the following command from the DC/OS CLI
+
+```
+$ sudo dcos tunnel vpn
+```
+
+The VPN client attempts to auto-configure DNS, but this functionality does not work on Mac OSX. To use the VPN client on OSX, [add the DNS servers](https://support.apple.com/kb/PH18499?locale=en_US) that DC/OS Tunnel instructs you to use.
+
+The VPN does not work by port forwarding, so you do not need to append `.mydcos.directory` to the end of your commands. For example, to access the DC/OS web interface from within the cluster, point your browser to `https://master.mesos`.
```
nlsun: Maybe we can remove this now that adding `.mydcos.directory` is a minority and only related to the HTTP proxy.

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,118 @@
...
+$ dcos tunnel http --port 8000
+```
+
+Then, configure your application to run HTTP on the port you specified above.
+
+### Port Forwarding
+The HTTP proxy works by port forwarding. Append `.mydcos.directory` to the end of your domain when you enter commands. For instance, `http://example.com:8080/?query=hello` becomes `http://example.com.mydcos.directory:8080/?query=hello`.
+
+### SRV Records
+The HTTP proxy exposes DC/OS SRV records as URLs in the form `_<port-name>._<service-name>._tcp.marathon.mesos.mydcos.directory`.
```
nlsun: I think I might have made an earlier comment about this, but the SRV record does not include `.mydcos.directory`, it's only used in the HTTP proxy because of the constraint that `.mydcos.directory` is appended to all URLs.

1.8/administration/access-node/tunnel.md
sascala: 👍

1.8/administration/access-node/tunnel.md
sascala: 👍

1.8/administration/access-node/tunnel.md
sascala: 👍

1.8/administration/access-node/tunnel.md
sascala: got it. I removed it above, but not here.

nlsun: Ok, sorry, I read this wrong, I thought this was the general "what is an SRV record" one but that was elsewhere. Since the purpose of this is to tell people "how to use SRV records with the HTTP proxy" then it should actually include `.mydcos.directory`. So the original version of this was correct.

sascala: indeed

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,116 @@
+---
+nav_title: Access by Proxy
+post_title: Access by Proxy using DC/OS Tunnel
+menu_order: 10
+---
+
+When developing applications on DC/OS, you may find it helpful to access your cluster from your local machine via SOCKS proxy, HTTP proxy, or VPN. For instance, you can work from your own development environment and immediately test against your DC/OS cluster.
+
+# SOCKS
+DC/OS Tunnel can run a SOCKS proxy over SSH to the cluster. SOCKS proxies work for any protocol, but your application must be configured to use the proxy, which runs on port 1080 by default. Access URLs by appending the `mydcos.directory` domain.
```
nlsun: the part about `mydcos.directory` should be removed

sascala: good catch!

nlsun:
LGTM

sascala:
🍾

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,116 @@
...
+A SRV DNS record is a mapping from a name to a IP/port pair. DC/OS creates SRV records in the form `_<port-name>._<service-name>._tcp.marathon.mesos`. The HTTP proxy exposes these as URLs. This feature can be useful for communicating with DC/OS services.
+
+# VPN
+DC/OS Tunnel provides you with full access to the DNS, masters, and agents from within the cluster. OpenVPN requires root privileges to configure these routes.
+
+# Using DC/OS Tunnel
+
+## Prerequisites
+* DC/OS 1.8+
+* Docker
```
joel-hamill: Why is Docker is listed here since this is already a prereq for using/installing DC/OS.

CC: @nlsun

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,116 @@
+---
+nav_title: Access by Proxy
```
joel-hamill: Doesn't tunnel support using both a proxy and VPN?

sascala: That's a good point. I was thinking of a VPN as a kind of proxy. But that may not be accurate. Thoughts, @nlsun ?

1.8/administration/access-node/tunnel.md
nlsun: I heard that Docker may not be required in the future (and become optional) because there will be a DC/OS container runtime.

1.8/administration/access-node/tunnel.md
nlsun: Hmm, I tried to come up with a term that captures both a proxy and vpn but I can't really think of one. Maybe just change this to "Proxy/VPN"?

1.8/administration/access-node/tunnel.md
joel-hamill: think we can remove and just say "DC/OS 1.8"

1.8/administration/access-node/tunnel.md
sascala: 👍

sascala:
@joel-hamill I changed the title to "Proxy and VPN" and if Nick's point about Docker seems to be a good one, I think this is ready to be merged.

1.8/administration/access-node/tunnel.md
sascala: In that case, we could remove the prereqs entirely, since these are the DC/OS docs. Having the CLI installed is a prerequisite, though. I'll add that and remove everything else.

sascala: (done)

nlsun:
Ok, another clarification (this one is a pretty big one I missed). The `.mydcos.directory` is only needed with the HTTP proxy when you use it on port 80 with no application modifications. This is when it operates in "transparent mode". This mode has the additional restriction that you cannot specify a different port.

If you configure your application to use the HTTP proxy (like what you'd do with the SOCKS proxy), then you you don't need `.mydcos.directory` and you can also change the port you want to access.

1.8/administration/access-node/tunnel.md
```diff
@@ -4,21 +4,108 @@ post_title: Access by Proxy and VPN using DC/OS Tunnel
...
+In transparent mode, the HTTP proxy runs as superuser on port 80 and does not require modification to your application. Access URLs by appending the `mydcos.directory` domain. You can also [use DNS SRV records as if they were URLs](#srv).
+
+## Standard Mode
+You must configure your service to use the HTTP proxy in standard mode, though it does not have any of the limitations of transparent mode. As in transparent mode, you can use [DNS SRV](#srv) records as URLs.
 
 <a name="srv"></a>
 ## SRV Records
 A SRV DNS record is a mapping from a name to a IP/port pair. DC/OS creates SRV records in the form `_<port-name>._<service-name>._tcp.marathon.mesos`. The HTTP proxy exposes these as URLs. This feature can be useful for communicating with DC/OS services.
 
+**Note:** The HTTP proxy cannot currently access HTTPS.
```
nlsun: This only occurs with transparent mode

1.8/administration/access-node/tunnel.md
```diff
@@ -4,21 +4,108 @@ post_title: Access by Proxy and VPN using DC/OS Tunnel
...
-DC/OS Tunnel can run a SOCKS proxy over SSH to the cluster. SOCKS proxies work for any protocol, but your application must be configured to use the proxy, which runs on port 1080 by default.
+DC/OS Tunnel can run a SOCKS proxy over SSH to the cluster. SOCKS proxies work for any protocol, but your service must be configured to use the proxy, which runs on port 1080 by default.
 
 # HTTP
-Run the HTTP proxy on your local port 80 to tunnel HTTP requests without modifying your application. Access URLs by appending the `mydcos.directory` domain. You can also [use DNS SRV records as if they were URLs](#srv). The HTTP proxy requires root access to use port 80, but can be configured to use a different port.
+
+The HTTP proxy can run in two modes: transparent and standard.
+
+##  Transparent Mode
+In transparent mode, the HTTP proxy runs as superuser on port 80 and does not require modification to your application. Access URLs by appending the `mydcos.directory` domain. You can also [use DNS SRV records as if they were URLs](#srv).
```
nlsun: Maybe mention that you cannot specify a port in the url (e.g. the `:5555` in `http://example.com:5555` won't work).

1.8/administration/access-node/tunnel.md
```diff
@@ -4,21 +4,108 @@ post_title: Access by Proxy and VPN using DC/OS Tunnel
...
+          <td>
+          <ul>
+              <li>SRV as URL</li>
+              <li>Specify ports</li>
+          </ul>
+          </td>
+          <td>
+              <ul>
+                  <li>Requires application configuration</li>
+                  <li>Only supports HTTP</li>
```
nlsun: Standard mode does HTTP/HTTPS

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,205 @@
+---
+nav_title: Access by Proxy and VPN
```
nlsun: Can we make it SUPER clear somewhere that this for development/debug/testing only? Never ever for production. Never.

sascala: Sure. Can you tell me a little about why this is not appropriate for production?

1.8/administration/access-node/tunnel.md
sascala: I added that information in the Usage section.

1.8/administration/access-node/tunnel.md
nlsun: Poor performance, mostly that it isn't designed for production use. Most people would assume that proxies and vpns are not for production usage but this is just for that 1 person who would try such a thing.

sascala: OK -- thanks! If it's a performance issue (and not, say, a security issue) I think I can just say not use it in production, rather than explain why.

1.8/administration/access-node/tunnel.md
```diff
@@ -79,7 +79,6 @@ DC/OS Tunnel provides you with full access to the DNS, masters, and agents from
           <td>
               <ul>
                   <li>Requires application configuration</li>
-                  <li>Only supports HTTP</li>
```
nlsun: I don't this should be removed, only changed to HTTP/HTTPS, since it doesn't support other protocols.

sascala: 👍

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,205 @@
...
+
+# HTTP
+
+The HTTP proxy can run in two modes: transparent and standard.
+
+##  Transparent Mode
+In transparent mode, the HTTP proxy runs as superuser on port 80 and does not require modification to your application. Access URLs by appending the `mydcos.directory` domain. You can also [use DNS SRV records as if they were URLs](#srv). The HTTP proxy cannot currently access HTTPS in transparent mode.
+
+## Standard Mode
+You must configure your service to use the HTTP proxy in standard mode, though it does not have any of the limitations of transparent mode. As in transparent mode, you can use [DNS SRV](#srv) records as URLs.
```
nlsun: Maybe change first sentence to "Though you must..., it does not..."

sascala: done

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,205 @@
...
+      <th>HTTP (transparent)</th>
+      <td>
+      <ul>
+          <li>SRV as URL</li>
+          <li>No application configuration</li>
+      </ul>
+      </td>
+      <td>
+          <ul>
+              <li>Cannot specify port</li>
```
nlsun: Cannot specify ports (expect through SRV)

1.8/administration/access-node/tunnel.md
```diff
@@ -0,0 +1,205 @@
...
+      <th>HTTP (transparent)</th>
+      <td>
+      <ul>
+          <li>SRV as URL</li>
+          <li>No application configuration</li>
+      </ul>
+      </td>
+      <td>
+          <ul>
+              <li>Cannot specify port (except through SRV)</li>
```
nlsun: port to ports, for uniformity with the other boxes

nlsun (commented):
Ok, that's the last thing I see (for now)

sascala:
@joel-hamill I made some more edits based on Nick's comments; I believe we're ready to go.

nlsun:
@joel-hamill ^

PR #597, opened 2016-09-15
updated unicon naming

sascala:


joel-hamill:
@sascala merge conflict

PR #601, opened 2016-09-16
changed overlay networks to virtual networks

sascala:
@somikbehera Could you please check this over? There were several places where I thought I should change "overlay" to "virtual" and I'd like to make sure that everything still reads well from a technical perspective.


joel-hamill:
Are there redirects setup for both sites? Any changes for enterprise-only content?

sascala:
good questions -- will check. Thank you.

PR #606, opened 2016-09-20
incorrect command for installing as public agent

liqdfire:
Corrected the documentation regarding changing the agent node type to a public agent. The command listed was how to see which nodes are public in the DCOS cli, not the bash command for running the install as a public node


joel-hamill:
Thanks!

PR #607, opened 2016-09-20
HAProxy in front of Admin Router

joel-hamill:


1.8/usage/haproxy.md
```diff
@@ -0,0 +1,77 @@
+---
+post_title: Configuring HTTP Proxy in Front of DC/OS
```
jgehrcke: "Configuring an HTTP Proxy ..." or "Configuring HAProxy ..."

("HAProxy" is a name, whereas "HTTP proxy" really is just a noun, so we need an article in the latter case)

jgehrcke: I think we should say "in front of DC/OS Admin Router" instead of just "in front of DC/OS". Then it's maybe clearer how this proxy setup differs from Marathon LB.

1.8/usage/haproxy.md
```diff
@@ -0,0 +1,77 @@
...
+nav_title: Configure HTTP Proxy
```
jgehrcke: this then needs to be adjusted to the new post_title

1.8/usage/haproxy.md
```diff
@@ -0,0 +1,77 @@
...
+---
+
+You can configure a secure communication channel between HAProxy and the DC/OS Admin Router, using a custom hostname and port.
+
+
+1.  Install HAProxy [1.6.9](http://www.haproxy.org/#down).
+
+1.  Create an HAProxy configuration for DC/OS. This example is for a DC/OS cluster on AWS. For more information on HAProxy configuration parameters, see the [documentation](https://cbonte.github.io/haproxy-dconv/configuration-1.6.html#3).
+
+    ```json
```
jgehrcke: This is not JSON (and also not YAML). Can we say just 'text' to prevent wrong syntax highlighting?

1.8/usage/haproxy.md
```diff
@@ -0,0 +1,77 @@
...
+    #---------------------------------------------------------------------
+    # Global settings
+    #---------------------------------------------------------------------
```
jgehrcke: This header can go (no value :-)).

1.8/usage/haproxy.md
```diff
@@ -0,0 +1,77 @@
...
+      timeout server  50000
+    
+    frontend http
+      # Bind on port 9090. HAProxy will listen on port 9090 on each 
+      # available network for new HTTP connections.
+      bind 0.0.0.0:9090
+      # Name of backend configuration for DC/OS.
+      default_backend dcos
+    
+      # Store request Host headers temporarily in transaction scope
```
jgehrcke: `headers` -> `header`

1.8/usage/haproxy.md
```diff
@@ -0,0 +1,77 @@
...
+      # Rewrite response Location header if it contains an absolute URL
+      # pointing to the 'dcoshost' host: replace 'dcoshost' with original
+      # request Host header (containing hostname and port).
+      http-response replace-header Location https?://dcoshost((/.*)?) "http://%[var(txn.request_host_header)]\1"
+    ```
+
+1.  Start, restart, or reload start HAProxy with these settings. 
+
+
+About the Location response header rewrite:
```
jgehrcke: We can leave this whole section out. It's good to have this tracked internally in DCOS-9559, but for the public docs this is in my opinion too much information.

1.8/usage/haproxy.md
```diff
@@ -0,0 +1,77 @@
...
+    global
+      daemon
+      log 127.0.0.1 local0
+      log 127.0.0.1 local1 notice
+      maxconn 20000
+      pidfile /var/run/haproxy.pid
+    #---------------------------------------------------------------------
+    # Common defaults that all the 'frontend' and 'backend' sections will 
+    # use if not designated in their block
+    #---------------------------------------------------------------------
```
jgehrcke: This header can go, too.

1.8/usage/haproxy.md
```diff
@@ -0,0 +1,77 @@
...
+      # Overwrite Host header to 'dcoshost'. This makes the Location
+      # header in DC/OS Admin Router upstream responses contain a
+      # predictable hostname (nginx uses this header value when
+      # constructing absolute redirect URLs). That value is used
+      # in the response Location header rewrite logic (see regular
+      # expression-based rewrite in the backend section below).
+      http-request set-header Host dcoshost
+    
+    backend dcos
+      # Use the source strategy for distributing load amongst the servers.
```
jgehrcke: This comment should go, because "for distributing load amongst the servers." raises the expectation to actually see multiple servers defined (which this cfg does not do).

1.8/usage/haproxy.md
```diff
@@ -0,0 +1,77 @@
+---
+post_title: Configuring HTTP Proxy in Front of DC/OS
+nav_title: Configure HTTP Proxy
+menu_order: 61
+---
+
+You can configure a secure communication channel between HAProxy and the DC/OS Admin Router, using a custom hostname and port.
```
jgehrcke: Let me try to find an introduction that describes the purpose of this a little more exhaustively...

"""
Sometimes it is required to put an HTTP proxy in front of DC/OS Admin Router. For that setup to work properly, the proxy needs to perform on-the-fly HTTP request and response header modification (the reason for that is that DC/OS is not aware of the custom hostname and port that is being used by user agents to address the HTTP proxy).

Below, you can find a tested HAProxy configuration example that takes care of named request/response rewriting. Furthermore, that configuration example ensures that the communication between HAProxy and DC/OS Admin Router is TLS-encrypted. It uses DC/OS' CA certificate so that HAProxy can verify the TLS certificate presented by Admin Router.
"""

1.8/usage/haproxy.md
```diff
@@ -0,0 +1,77 @@
...
+      # predictable hostname (nginx uses this header value when
+      # constructing absolute redirect URLs). That value is used
+      # in the response Location header rewrite logic (see regular
+      # expression-based rewrite in the backend section below).
+      http-request set-header Host dcoshost
+    
+    backend dcos
+      # Use the source strategy for distributing load amongst the servers.
+      balance source
+      # DC/OS CA cert downloaded from dcoshost/ca/dcos-ca.crt
```
jgehrcke: Let's change this line (the goal is to add more clarity):

```
# The local file `dcos-ca.crt` is the root CA certificate corresponding to
# that individual DC/OS cluster that is being proxied to. It must be retrieved
# out-of-band, e.g. via https://dcoshost/ca/dcos-ca.crt
```

1.8/usage/haproxy.md
```diff
@@ -0,0 +1,77 @@
...
+      server dcos-1 52.57.1.99:443 check-ssl ssl verify required ca-file dcos-ca.crt verifyhost frontend-ElasticL-JW6HOE0W6MAT-1017800469.eu-central-1.elb.amazonaws.com
+    
+      # Rewrite response Location header if it contains an absolute URL
+      # pointing to the 'dcoshost' host: replace 'dcoshost' with original
+      # request Host header (containing hostname and port).
+      http-response replace-header Location https?://dcoshost((/.*)?) "http://%[var(txn.request_host_header)]\1"
+    ```
+
+1.  Start, restart, or reload start HAProxy with these settings.
```
jgehrcke: Two times "start" :). I think it's fine to just say "Start HAProxy with these settings."

jgehrcke (changes requested):
Thanks, Joel! I have left comments.

1.8/usage/haproxy.md
joel-hamill: @jgehrcke is this HAProxy example only for enterprise?

jgehrcke: Well, no, it's generally valid for Open DC/OS, too. However, certificate verification will not work, because there is no CA certificate to verify against (unless people inject their own certificate for Admin Router). 

Should we provide an alternative (outcommented) upstream server definition line that will not require certificate verification?

CC @alberts

joel-hamill: sounds like we'll need 2 versions of this: enterprise and DC/OS.

1.8/usage/service-discovery/haproxy-adminrouter.md
```diff
@@ -0,0 +1,69 @@
...
+      # in the response Location header rewrite logic (see regular
+      # expression-based rewrite in the backend section below).
+      http-request set-header Host dcoshost
+    
+    backend dcos
+      balance source
+      # The local file `dcos-ca.crt` is the root CA certificate that corresponds to
+      # that individual DC/OS cluster that is being proxied to. It must be retrieved
+      # out-of-band, for example via https://dcoshost/ca/dcos-ca.crt.
+      server dcos-1 52.57.1.99:443 check-ssl ssl verify required ca-file dcos-ca.crt verifyhost frontend-ElasticL-JW6HOE0W6MAT-1017800469.eu-central-1.elb.amazonaws.com
```
jgehrcke: I gave this more thought. The instructions here hold true for both, Enterprise DC/OS as well as Open DC/OS (the only difference being where to obtain the CA certificate from, which is a problem more-or-less unrelated to the problem that this docs page addresses). To make this clear, I worked on a more exhaustive commentary explaining what the `server ....` line is actually doing. What I would like to see changed here now:

1) remove the `balance source` (it is a distraction, as it is unrelated to the problem addressed in this docs page, and people will pimp their HAProxy config anyway beyond what's shown here).

2) Replace the current `server ...` line _and_ its commentary (three lines) with the following:
```
  # Variant 1: use TLS-encrypted communication with DC/OS Admin Router and
  # perform server certificate verification (including hostname verification).
  # Unless you are using Mesosphere Enterprise DC/OS, this requires you to
  # configure Admin Router with a custom TLS server certificate, see
  # https://dcos.io/docs/1.8/administration/securing-your-cluster/.
  #
  # Explanation for the parameters in the following `server` definition line:
  # 
  # 1.2.3.4:443
  #
  #   IP address and port that HAProxy uses to connect to DC/OS Admin
  #   Router. This needs to be adjusted to your setup.
  #   
  #
  # ssl verify required
  #
  #   Instruct HAProxy to use TLS, and to error out if server certificate
  #   verification fails.
  #
  # ca-file dcos-ca.crt
  #
  #   The local file `dcos-ca.crt` is expected to contain the CA certificate
  #   that Admin Router's certificate will be verified against. It must be
  #   retrieved out-of-band (on Mesosphere Enterprise DC/OS this can be
  #   obtained via https://dcoshost/ca/dcos-ca.crt)
  #
  # verifyhost frontend-xxx.eu-central-1.elb.amazonaws.com
  #
  #   When verifying the TLS certificate presented by DC/OS Admin Router,
  #   perform hostname verification using the hostname specified here
  #   (expect the server certificate to contain a DNSName SAN that is
  #   equivalent to the hostname defined here). The hostname shown here is
  #   just an example and needs to be adjusted to your setup.

  server dcos-1 1.2.3.4:443 ssl verify required ca-file dcos-ca.crt verifyhost frontend-xxx.eu-central-1.elb.amazonaws.com

  # Variant 2: use TLS-encrypted communication with DC/OS Admin Router, but do
  # not perform server certificate verification (warning: this is insecure, and
  # we hope that you know what you are doing).
  
  # server dcos-1 1.2.3.4:443 ssl verify none
```

1.8/usage/haproxy.md
jgehrcke: The instructions here hold true for both, Enterprise DC/OS as well as Open DC/OS (the only difference being where to obtain the CA certificate from, which is a problem more-or-less unrelated to the problem that this docs page here is supposed to address). In order to reduce maintenance efforts, let's try to not split this up. I submit another review for suggesting corresponding changes.

1.8/usage/haproxy.md
jgehrcke: I don't see this addressed.

1.8/usage/haproxy.md
jgehrcke: In your changeset, you have introduced this sentence:

> You can configure a secure communication channel from a custom hostname and port to DC/OS Admin Router by using an HTTP Proxy.

"from a custom hostname and port to DC/OS Admin Router" does not make sense, technically. 

Can we, like I proposed, write something like "Sometimes it is required to put an HTTP proxy in front of DC/OS Admin Router." to motivate this docs page? The point is that we do _not know_ why customers want to place an HTTP proxy in front of Admin Router. They may have various good (or bad :-)) reasons. The point of this docs page really is: _if_ they need to do so (for whatever reason), then we can help them.

1.8/usage/service-discovery/haproxy-adminrouter.md
```diff
@@ -0,0 +1,69 @@
...
+post_title: Configuring HTTP Proxy in Front of Admin Router
+nav_title: Configure HTTP Proxy
+menu_order: 61
+---
+
+You can configure a secure communication channel from a custom hostname and port to DC/OS Admin Router by using an HTTP Proxy.
+ 
+The HTTP Proxy must perform on-the-fly HTTP request and response header modification because DC/OS is not aware of the custom hostname and port that is being used by user agents to address the HTTP proxy.
+
+These instructions provide a tested [HAProxy](http://www.haproxy.org/) configuration example that handles the named request/response rewriting. This example ensures that the communication between HAProxy and DC/OS Admin Router is TLS-encrypted. It uses the DC/OS [CA certificate](/1.8/administration/tls-ssl/) so that HAProxy can verify the TLS certificate presented by Admin Router.
```
jgehrcke: A reasonable simplification that renders this introduction non-specific for Open or EE: we can remove the last sentence here.

That is, change
```
This example ensures that [...] is TLS-encrypted. It uses the DC/OS [CA certificate](/1.8/administration/tls-ssl/) so that HAProxy can verify the TLS certificate presented by Admin Router.
```
to
```
This example ensures that [...] is TLS-encrypted.
```

jgehrcke (changes requested):
I have added a 2nd review that proposes changes for

1) doing a much better job explaining the `server ...` line in the config.
2) making this docs page apply to both Open DC/OS as well as Enterprise DC/OS.

1.8/usage/service-discovery/haproxy-adminrouter.md
joel-hamill: Is TLS-encrypted communication included out-of-the-box with Enterprise?

joel-hamill:
@jgehrcke i have updated per your comments - and left a question above. thanks.

1.8/usage/service-discovery/haproxy-adminrouter.md
jgehrcke: I read your question like this: "Is Variant 1 above supported on a fresh and default Mesosphere Enterprise DC/OS 1.8 install?" and can answer with "yes" :)

Note that if taken literally, "TLS-encrypted communication" is also included out-of-the-box with Open DC/OS. The significant differences really are in the details around X.509 certificate management.

jgehrcke:
@joel-hamill Thanks, I have answered your question!

Would be great if you could explicitly ack or disregard these two comments:

* https://github.com/dcos/dcos-docs/pull/607#discussion_r80022666
* https://github.com/dcos/dcos-docs/pull/607#discussion_r80021886

1.8/usage/haproxy.md
joel-hamill: The original proposed sentence is a bit ambiguous. I'd propose `You can put an HTTP proxy in front of the DC/OS Admin Router.`

joel-hamill:
@jgehrcke is this ready to ship?

1.8/usage/service-discovery/haproxy-adminrouter.md
```diff
@@ -0,0 +1,107 @@
+---
+post_title: Configuring HAProxy in Front of Admin Router
+nav_title: Configure HAProxy
+menu_order: 61
+---
+
+You can use HAPRoxy to set up an HTTP proxy in front of the DC/OS Admin Router.
```
jgehrcke: "HAPRoxy" -> HAProxy

1.8/usage/haproxy.md
jgehrcke: Note just for me: now addressed.

1.8/usage/haproxy.md
jgehrcke: That's great, yes!

jgehrcke:
I have added one more tiny thing to change. LGTM otherwise, thanks @joel-hamill!

1.8/usage/service-discovery/haproxy-adminrouter.md
```diff
@@ -0,0 +1,107 @@
+---
+post_title: Configuring HAProxy in Front of Admin Router
+nav_title: HAProxy in Front of Admin Router
+menu_order: 400
+---
+
+You can use HAProxy to set up an HTTP proxy in front of the DC/OS [Admin Router](/docs/1.8/overview/concepts/#adminrouter).
```
emanic: What is the reason for doing this? Might be helpful to discuss that. I am not sure, but I think this is to prevent warnings from self-signed certificates? But perhaps there are other reasons to do this. It might be good to discuss the reasons for doing this. If I am correct and it is for the self-signed certificate issue, then I can add a link to this page from my other section. I am guessing that the discussion of this issue could be somewhat complex because we want to single-source this content from open to enterprise. I am not super familiar with the open source offering, but I believe that the use case is somewhat similar there, where they have a "snake-oil certificate", but I am not up on the details at this time.

jgehrcke (approved):

1.8/usage/service-discovery/haproxy-adminrouter.md
emanic: @joel-hamill It varies in Enterprise based on the security mode, as discussed in the TLS page

1.8/usage/service-discovery/haproxy-adminrouter.md
```diff
@@ -0,0 +1,107 @@
...
+      #
+      #   Instruct HAProxy to use TLS, and to error out if server certificate
+      #   verification fails.
+      #
+      # ca-file dcos-ca.crt
+      #
+      #   The local file `dcos-ca.crt` is expected to contain the CA certificate
+      #   that Admin Router's certificate will be verified against. It must be
+      #   retrieved out-of-band (on Mesosphere Enterprise DC/OS this can be
+      #   obtained via https://dcoshost/ca/dcos-ca.crt)
```
emanic: We discuss how to retrieve this in the Enterprise docs on the TLS page, but not sure if it is the same for open.

1.8/usage/service-discovery/haproxy-adminrouter.md
```diff
@@ -0,0 +1,107 @@
...
+      retries             3
+      maxconn          20000
+      timeout connect  5000
+      timeout client  50000
+      timeout server  50000
+    
+    frontend http
+      # Bind on port 9090. HAProxy will listen on port 9090 on each 
+      # available network for new HTTP connections.
+      bind 0.0.0.0:9090
```
alberts: somewhere here we should add a directive to use a different server certificate

1.8/usage/service-discovery/haproxy-adminrouter.md
jgehrcke: > What is the reason for doing this? Might be helpful to discuss that.

There is a plethora of reasons. Some motivation is performance-related, some reasons may be policy-guided, other reasons may not even be good reasons, and other reasons are security-related in the widest sense. Overall, there are so many reasons (most of them unrelated to certificates), that I wouldn't go into mentioning specific reasons here on that page.

> If I am correct and it is for the self-signed certificate issue, then I can add a link to this page from my other section.

This help page was actually written with a motivation quite different from certificates. But yes, the solution we describe here can also be used to let someone take full control over the certificates presented to web browsers when accessing DC/OS via HTTPS through the proxy.

So, having a link from somewhere else discussing this certificate-related issue _to_ this help page here sounds perfectly fine.

1.8/usage/service-discovery/haproxy-adminrouter.md
alberts: bind *:443 ssl crt /path/to/browser-trusted.crt

1.8/usage/service-discovery/haproxy-adminrouter.md
jgehrcke: Does not currently apply to Open DC/OS.

1.8/usage/service-discovery/haproxy-adminrouter.md
emanic: @jgehrcke Thanks! I was talking to Albert about this because I was not sure where one would insert the certificate that is signed by the valid certificate authority...I didn't see it here. Do you think it might be good to add? He was saying there were two use cases so far that we have run into: to run on a custom port and to use a valid cert. We could possibly mention those as "for example," I don't know if you might be amenable. Just as a couple of examples to help people who might not be sure, let me know your thoughts.

1.8/usage/service-discovery/haproxy-adminrouter.md
joel-hamill: Added as 
```    frontend http
      # Bind on port 9090. HAProxy will listen on port 9090 on each 
      # available network for new HTTP connections.
      bind 0.0.0.0:9090
      # Specify a different server certificate.
      #
      # bind *:9091 ssl crt /path/to/browser-trusted.crt
      # Name of backend configuration for DC/OS.
      default_backend dcos```

File: 1.8/usage/service-discovery/haproxy-adminrouter.md (hunk 16 above)
Comment (emanic) [inline 1.8/usage/service-discovery/haproxy-adminrouter.md]: It might be a good thing to mention somewhere that this section is applicable to Enterprise only. We could also have different content on open versus Enterprise, but I can see how that might be less efficient.

Comment (jgehrcke) [inline 1.8/usage/service-discovery/haproxy-adminrouter.md]: I should have responded more exhaustively to your original comment.

> not sure if it is the same for open.

This is described a few lines above:

```
+      # If you are using the community-supported version of DC/OS, you must 
+      # configure Admin Router with a custom TLS server certificate, see 
+      # https://dcos.io/docs/1.8/administration/securing-your-cluster/.
```

Also, I meant to say "`on Mesosphere Enterprise DC/OS this can be obtained via https://dcoshost/ca/dcos-ca.crt` does not curently apply to Open DC/OS" (as hopefully is apparent from the sentence itself).

> It might be a good thing to mention somewhere that this section is applicable to Enterprise only. 

Hope it's clear now that this is not true. The configuration shown is valid for both, literally. What's different is _how_ to obtain `dcos-ca.crt`. This is covered for both scenarios 
* via linking to other content in case of Open DC/OS and
* via a short description in case of Enterprise.

Ok?

Comment (emanic) [inline 1.8/usage/service-discovery/haproxy-adminrouter.md]: Oh, OK. Sounds good! 👍 Thanks JP...

File: 1.8/usage/service-discovery/haproxy-adminrouter.md (hunk 15 above)
Comment (jgehrcke) [inline 1.8/usage/service-discovery/haproxy-adminrouter.md]: > let me know your thoughts

Well, this is IMO low priority and not risk-free. So, let's save this for later.

File: 1.8/usage/service-discovery/haproxy-adminrouter.md (hunk 17 above)
Comment (jgehrcke) [inline 1.8/usage/service-discovery/haproxy-adminrouter.md]: @joel-hamill @alberts Here I was missing how to specify the private key (which is required for configuring a server certificate) -- so I looked up the docs again, and it explains all the magic of the `crt` parameter here. I think we should add the documentation link here and slightly change wording. Can we change this block:
```
      # Specify a different server certificate.
      #
      # bind *:9091 ssl crt /path/to/browser-trusted.crt
      # Name of backend configuration for DC/OS.
```

to:
```
      # Specify your own server certificate chain and associated private key.
      # See https://cbonte.github.io/haproxy-dconv/configuration-1.6.html#5.1-crt
      # bind *:9091 ssl crt /path/to/browser-trusted.crt
      #
      # Name of backend configuration for DC/OS.
```

jgehrcke (changes requested):

1.8/usage/service-discovery/haproxy-adminrouter.md
joel-hamill: added via this PR #656

PR #611, opened 2016-09-21
Update 1.8 proxy config docs to use new path

BenWhitehead:
In dc/os 1.8 proxy configuration for cosmos was changed to use
`/opt/mesosphere/etc/proxy.env` instead of `/var/lib/dcos/environment.proxy`.
The docs around how to configure proxy support for cosmos are updated to the
new path.


BenWhitehead:
Closing in favor of fix that will be made for https://dcosjira.atlassian.net/browse/DCOS-398

PR #614, opened 2016-09-22
Add general warning about UI and APIs during upgrade

branden:


1.7/administration/upgrading.md
```diff
@@ -11,6 +11,7 @@ This document provides instructions for upgrading a DC/OS cluster from version 1
 
 - The Advanced Installation method is the _only_ recommended upgrade path for DC/OS. It is recommended that you familiarize yourself with the [Advanced DC/OS Installation Guide][advanced-install] before proceeding.
 - The [VIP features](/docs/1.7/usage/service-discovery/virtual-ip-addresses/), added in DC/OS 1.7, require that ports 32768 - 65535 are open between all agent and master nodes for both TCP and UDP.
+- The DC/OS UI and APIs may be inconsistent or unavailable while masters are being upgraded. Avoid using them until all masters have been upgraded and have rejoined the cluster. You can monitor the health of a master during an upgrade by watching Exhibitor on port 8181.
```
adam-mesos: or `/exhibitor` instead of port 8181

branden: The Adminrouter dependency there is why I went with the direct Exhibitor port. If any of `ready.spartan`, `leader.mesos`, or `marathon.mesos` don't resolve, then the Adminrouter won't come up and you won't be able to reach `/exhibitor`. So it's not as useful for watching an in-flight upgrade.

PR #630, opened 2016-09-27
Upgrade doc update

branden:
* Upgrade ZooKeeper leader last (1.7 and 1.8)
* Validate upgrade with Mesos metrics (1.8)

Pinging @cmaloney and @joel-hamill for review.

### DIFF
--- a/1.7/administration/upgrading.md
+++ b/1.7/administration/upgrading.md
@@ -51,7 +51,13 @@ This document provides instructions for upgrading a DC/OS cluster from version 1
 
 ### DC/OS Masters
 
-Identify your Mesos leader node. This node should be the last master node that you upgrade. Proceed with upgrading every master node using the following procedure. When you complete each upgrade, monitor the logs to ensure the unit has re-joined the cluster and completed reconciliation.
+Identify the ZooKeeper leader among the masters. This node should be the last master node that you upgrade. You can determine whether a master node is a ZooKeeper leader by sending the `stat` command to the ZooKeeper client port.
+
+```
+$ echo stat | /opt/mesosphere/bin/toybox nc localhost 2181 | grep "Mode:"
+```
+
+Proceed with upgrading every master node using the following procedure. When you complete each upgrade, monitor the logs to ensure the node has re-joined the cluster and completed reconciliation.
 
 1.  Download the `dcos_install.sh` script:
 
--- a/1.7/administration/upgrading.md
+++ b/1.7/administration/upgrading.md
@@ -57,7 +57,7 @@ Identify the ZooKeeper leader among the masters. This node should be the last ma
 $ echo stat | /opt/mesosphere/bin/toybox nc localhost 2181 | grep "Mode:"
 ```
 
-Proceed with upgrading every master node using the following procedure. When you complete each upgrade, monitor the logs to ensure the node has re-joined the cluster and completed reconciliation.
+Proceed with upgrading every master node using the following procedure. When you complete each upgrade, monitor the logs to ensure the node has rejoined the cluster and completed reconciliation.
 
 1.  Download the `dcos_install.sh` script:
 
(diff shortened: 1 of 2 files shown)


1.7/administration/upgrading.md
```diff
@@ -51,7 +51,13 @@ This document provides instructions for upgrading a DC/OS cluster from version 1
 
 ### DC/OS Masters
 
-Identify your Mesos leader node. This node should be the last master node that you upgrade. Proceed with upgrading every master node using the following procedure. When you complete each upgrade, monitor the logs to ensure the unit has re-joined the cluster and completed reconciliation.
+Identify the ZooKeeper leader among the masters. This node should be the last master node that you upgrade. You can determine whether a master node is a ZooKeeper leader by sending the `stat` command to the ZooKeeper client port.
+
```
joel-hamill: Should we add which platform this example is for?

1.7/administration/upgrading.md
```diff
@@ -51,7 +51,13 @@ This document provides instructions for upgrading a DC/OS cluster from version 1
...
+```
+$ echo stat | /opt/mesosphere/bin/toybox nc localhost 2181 | grep "Mode:"
+```
+
+Proceed with upgrading every master node using the following procedure. When you complete each upgrade, monitor the logs to ensure the node has re-joined the cluster and completed reconciliation.
```
joel-hamill: `rejoined`

1.8/administration/upgrading.md
```diff
@@ -136,7 +143,8 @@ Identify your Mesos leader node. This node should be the last master node that y
 
 1.  Validate the upgrade
 
-    Monitor the Mesos UI to verify that the upgraded node re-joins the DC/OS cluster and that tasks are reconciled (`http://<dcos_master>/mesos`).
+    - Verify that `curl http://<dcos_agent_private_ip>:5051/metrics/snapshot` has the metric `slave/registered` with a value of `1`.
+    - Monitor the Mesos UI to verify that the upgraded node re-joins the DC/OS cluster and that tasks are reconciled (`http://<dcos_master>/mesos`).
```
joel-hamill: `rejoin`

1.8/administration/upgrading.md
```diff
@@ -136,7 +143,8 @@ Identify your Mesos leader node. This node should be the last master node that y
 
 1.  Validate the upgrade
 
-    Monitor the Mesos UI to verify that the upgraded node re-joins the DC/OS cluster and that tasks are reconciled (`http://<dcos_master>/mesos`).
+    - Verify that `curl http://<dcos_agent_private_ip>:5051/metrics/snapshot` has the metric `slave/registered` with a value of `1`.
+    - Monitor the Mesos UI to verify that the upgraded node rejoins the DC/OS cluster and that tasks are reconciled (`http://<dcos_master>/mesos`).
```
joel-hamill: Can this be done from the DC/OS UI?

1.7/administration/upgrading.md
branden: This assumes that the OS provides `echo` and `grep`, which are common utilities, and it doesn't use them with any platform-specific flags. This example should be portable among most distros, certainly all that we support. /cc @cmaloney

1.8/administration/upgrading.md
branden: It can, but the Mesos UI provides a more detailed view of connected agents and their tasks. It also doesn't change much between releases, so it's easier to reuse these instructions with upcoming DC/OS versions.

joel-hamill: I could see that...however, we're trying to gain DC/OS adoption/usage :) But ultimately it's your call.

cmaloney: For this for now I'd rather say Mesos UI, there are several behaviors of the DC/OS UI that work against what we want / need here currently.

1.7/administration/upgrading.md
```diff
@@ -51,7 +51,13 @@ This document provides instructions for upgrading a DC/OS cluster from version 1
...
 ### DC/OS Masters
 
-Identify your Mesos leader node. This node should be the last master node that you upgrade. Proceed with upgrading every master node using the following procedure. When you complete each upgrade, monitor the logs to ensure the unit has re-joined the cluster and completed reconciliation.
+Identify the ZooKeeper leader among the masters. This node should be the last master node that you upgrade. You can determine whether a master node is a ZooKeeper leader by sending the `stat` command to the ZooKeeper client port.
+
+```
+$ echo stat | /opt/mesosphere/bin/toybox nc localhost 2181 | grep "Mode:"
+```
+
+Proceed with upgrading every master node using the following procedure. When you complete each upgrade, monitor the logs to ensure the node has rejoined the cluster and completed reconciliation.
```
cmaloney: Are there any caveats around the leading mesos master / leading  marathon we want to note here? Those both need to go through reconiciliation when they are failed over.

All the old / completed tasks also disappear from the Mesos UI if I recall.

cmaloney:
Generally looks good to me, left comments

1.7/administration/upgrading.md
branden: Added a note about disappearing task history in the Mesos UI.

PR #744, opened 2016-11-08
service-discovery: Add an overview of DNS

nlsun:
- [ ] Bug fix
- [x] New feature 

- [ ] Blocker
- [ ] High
- [x] Medium

- [x] I have tested all commands and code snippets.
- [x] I have built locally and tested for broken links and formatting.
- [x] I have made this change for all affected versions (e.g. 1.7, 1.8, and 1.9).
- [x] My doc follows the style guide: https://github.com/dcos/dcos-docs#contributing.

### DIFF
--- a/1.8/usage/service-discovery/dns-overview.md
+++ b/1.8/usage/service-discovery/dns-overview.md
@@ -1,0 +1,51 @@
+---
+post_title: DNS Quick Reference
+nav_title: DNS Reference
+menu_order: 20
+---
+
+This quick reference provides a summary of the available options.
+
+To help explain, we'll use this imaginary application:
+* The Service is in the following hierarchy:
+ * Group: `outergroup` > Group: `subgroup` > Service Name: `myapp`
+* Port: `555`
+ * Port Name: `myport`
+ * Load Balanced
+* Running on a
+  [Virtual Network](/docs/1.8/usage/service-discovery/load-balancing-vips/overlay-networks/)
+* Running on the Marathon framework (if unsure, it's probably this one)
+ * If you are running another framework, then replace any instance of
+   `marathon` with the name of your framework.
+
+# Service Discovery Options
+
+Use one of these options to find the DNS name for your task.
+You should choose the first option that satisfies your requirements:
+
+1.  `outergroupsubgroupmyapp.marathon.l4lb.thisdcos.directory:555`
+ * This is only available when the service is load balanced. `:555` is not a
+   part of the DNS address, but is there to show that this address and port
+   is load balanced as a pair rather than individually.
+1.  `myapp-subgroup-outergroup.marathon.containerip.dcos.thisdcos.directory`
+ * This is only available when the service is running on a
+   [virtual network](/docs/1.8/usage/service-discovery/load-balancing-vips/overlay-networks/).
+1.  `myapp-subgroup-outergroup.marathon.agentip.dcos.thisdcos.directory`
+ * This is always available and should be used when the service is not
+   running on a
+   [virtual network](/docs/1.8/usage/service-discovery/load-balancing-vips/overlay-networks/).
+1.  `myapp-subgroup-outergroup.marathon.autoip.dcos.thisdcos.directory`
+ * This is always available and should be used to address an application that is
+   transitioning on or off a
+   [virtual network](/docs/1.8/usage/service-discovery/load-balancing-vips/overlay-networks/).
+1.  `myapp-subgroup-outergroup.marathon.mesos`
+ * This is always available, and is equivalent for the most part to the
+   `agentip`. However it is less specific and less performant than the
+   `agentip` and thus use is discouraged.
+
+Other discovery option(s):
+
+* `_myport._myapp.subgroup.outergroup._tcp.marathon.mesos`
+ * This is not a DNS A record but rather a DNS SRV record. This is only
+   available when the port has a name. SRV records are a mapping from a
+   name to an "Address + Port" pair.
--- a/1.9/usage/service-discovery/dns-overview.md
+++ b/1.9/usage/service-discovery/dns-overview.md
@@ -1,0 +1,51 @@
+---
+post_title: DNS Quick Reference
+nav_title: DNS Reference
+menu_order: 20
+---
+
+This quick reference provides a summary of the available options.
+
+To help explain, we'll use this imaginary application:
+* The Service is in the following hierarchy:
+ * Group: `outergroup` > Group: `subgroup` > Service Name: `myapp`
+* Port: `555`
+ * Port Name: `myport`
+ * Load Balanced
+* Running on a
+  [Virtual Network](/docs/1.9/usage/service-discovery/load-balancing-vips/overlay-networks/)
+* Running on the Marathon framework (if unsure, it's probably this one)
+ * If you are running another framework, then replace any instance of
+   `marathon` with the name of your framework.
+
+# Service Discovery Options
+
+Use one of these options to find the DNS name for your task.
+You should choose the first option that satisfies your requirements:
+
+1.  `outergroupsubgroupmyapp.marathon.l4lb.thisdcos.directory:555`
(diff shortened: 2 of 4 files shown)


1.8/usage/service-discovery/dns-overview.md
```diff
@@ -0,0 +1,68 @@
+---
+post_title: DNS Quick Reference
+nav_title: DNS Reference
+menu_order: 20
+---
+
+We've got a lot going on when it comes to DNS and this article aims to
```
joel-hamill: Suggest rewrite: `DC/OS offers a lot flexibility and extensibility for DNS. This quick reference provides a summary of the available options.`

1.9/usage/service-discovery/dns-overview.md
```diff
@@ -0,0 +1,68 @@
+---
+post_title: DNS Quick Reference
+nav_title: DNS Reference
+menu_order: 20
+---
+
+We've got a lot going on when it comes to DNS and this article aims to
+summarize all that information in a single place.
+
+# Example Setup
```
joel-hamill: Move this example to the bottom, after the other `# What there is`.

1.9/usage/service-discovery/dns-overview.md
```diff
@@ -0,0 +1,68 @@
...
+On this example setup:
+* Default DC/OS Name (if unsure, it's probably this one)
+ * If you named your DC/OS something else, replace any instance of
+   `thisdcos` with the name that you selected.
+
+Our imaginary `myapp` is a web server listening on port `555`.
+We'll be using `curl` as our client application.
+
+# What There Is
+
```
joel-hamill: Needs an intro. What is this a list of?

1.9/usage/service-discovery/dns-overview.md
nlsun: I think it helps to set the context first

1.9/usage/service-discovery/dns-overview.md
```diff
@@ -0,0 +1,68 @@
...
+menu_order: 20
+---
+
+We've got a lot going on when it comes to DNS and this article aims to
+summarize all that information in a single place.
+
+# Example Setup
+
+To make things more concrete, we'll use this example application:
+* (Group) `outergroup` > (Group) `innergroup` > (Service Name) `myapp`
```
nlsun: Do you mean JSON instead of the bulleted list? I think it's more concise to only specify the relevant aspects of the example service

1.9/usage/service-discovery/dns-overview.md
```diff
@@ -0,0 +1,68 @@
...
+   `marathon` with the name of your framework.
+
+On this example setup:
+* Default DC/OS Name (if unsure, it's probably this one)
+ * If you named your DC/OS something else, replace any instance of
+   `thisdcos` with the name that you selected.
+
+Our imaginary `myapp` is a web server listening on port `555`.
+We'll be using `curl` as our client application.
+
```
nlsun: Yeah, i'll remove the part about `curl`, it's not needed here.

1.9/usage/service-discovery/dns-overview.md
joel-hamill: yes, the JSON. i think it would help with context. without the reader is forced to envision the rest.

nlsun:
I've made changes to address the comments

1.9/usage/service-discovery/dns-overview.md
nlsun: Well this one is a reference sheet rather than a full tutorial, and the rest of the JSON would just be a distraction. I also want to keep this as short and to the point as possible.

nlsun:
I've made more changes, it's ready again.

1.8/usage/service-discovery/dns-overview.md
```diff
@@ -0,0 +1,65 @@
+---
+post_title: DNS Quick Reference
+nav_title: DNS Reference
+menu_order: 20
+---
+
+DC/OS offers a lot when it comes to DNS.
```
joel-hamill: I think this sentence can be removed entirely.

1.8/usage/service-discovery/dns-overview.md
```diff
@@ -0,0 +1,65 @@
...
+This quick reference provides a summary of the available options.
+
+# Example Setup
+
+To make things more concrete, we'll use this example application:
+* (Group) `outergroup` > (Group) `subgroup` > (Service Name) `myapp`
```
joel-hamill: What do these carrots mean: `>`?

1.8/usage/service-discovery/dns-overview.md
```diff
@@ -0,0 +1,65 @@
...
+* Port: `555`
+ * Port Name: `myport`
+ * Load Balanced
```
joel-hamill: Bullets from line 16 - 20 are not attributes of a reference app. These are general attributes/recommendations.

1.8/usage/service-discovery/dns-overview.md
```diff
@@ -0,0 +1,65 @@
...
+nav_title: DNS Reference
+menu_order: 20
+---
+
+DC/OS offers a lot when it comes to DNS.
+This quick reference provides a summary of the available options.
+
+# Example Setup
+
+To make things more concrete, we'll use this example application:
```
joel-hamill: Because there is not actual example, suggest rewrite: `To help explain concepts, this document uses a theoretical application for reference:`

1.8/usage/service-discovery/dns-overview.md
```diff
@@ -0,0 +1,65 @@
...
+* Port: `555`
+ * Port Name: `myport`
+ * Load Balanced
+* Running on a Virtual Network
+* Running on the Marathon framework (if unsure, it's probably this one)
+ * If you are running another framework, then replace any instance of
+   `marathon` with the name of your framework.
+
+On this example setup:
+* Default DC/OS Name (if unsure, it's probably this one)
```
joel-hamill: Replace lines 22 - 25 with: DC/OS uses `thisdcos` as the default DC/OS name. If you provided a custom DC/OS name, replace any instance of `thisdcos` with the name that you provided.

1.8/usage/service-discovery/dns-overview.md
nlsun: It's the group hierarchy, I can experiment with nesting bullets to see if it's clearer.

nlsun:
Ok, I've made changes.

joel-hamill:
@nlsun An example, or excerpt of an example would be worth a thousand words here. Please try this instead. Otherwise we are creating a whole other convention for people to possibly be confused by (these carrots). Start with an actual ref app and explain out from there.

nlsun:
In the JSON, it would look like `/outergroup/subgroup/myapp`. This is way more ambiguous than spelling out the fact that it's a group/group/service because someone might actually try to create an app with the name as such instead of realizing that it's a service inside groups.

If I were to do an example, then I'd also have to walk people through creating groups, and then a large piece of JSON for the actual application. This then quickly becomes a rabbit hole where we are presenting too much extraneous information and the actual purpose (which is very short) becomes hidden. That might be fine for a normal doc page because their purpose is to be a tutorial, but this is just a quick reference. If someone wants details they will go explore the rest of the docs.

1.8/usage/service-discovery/dns-overview.md
```diff
@@ -0,0 +1,61 @@
...
+menu_order: 20
+---
+
+This quick reference provides a summary of the available options.
+
+# Example Setup
+
+To help explain, we'll use this imaginary application:
+* The Service is in the following hierarchy:
+ * Group: `outergroup` > Group: `subgroup` > Service Name: `myapp`
```
joel-hamill: How about changing to `/outergroup/subgroup/myapp`?

nlsun:
Okay, finished this round of tweaks.

PR #812, opened 2016-12-26
Graceful shutdown info

joel-hamill:
- [ ] Bug fix
- [x] New feature 

- [ ] Blocker
- [ ] High
- [x] Medium

- [ ] I have tested all commands and code snippets.
- [ ] I have built locally and tested for broken links and formatting.
- [ ] I have made this change for all affected versions (e.g. 1.7, 1.8, and 1.9).
- [ ] My doc follows the style guide: https://github.com/dcos/dcos-docs#contributing.

## If you included a comment with your commit, it appears here:

### DIFF
--- a/1.8/administration/installing/custom/graceful-shutdown.md
+++ b/1.8/administration/installing/custom/graceful-shutdown.md
@@ -1,0 +1,12 @@
+---
+post_title: Graceful Node Shutdown
+menu_order: 501
+---
+
+You can use these instructions to remove a DC/OS cluster node.
+
+1.  Add a [Mesos maintenance primitive](http://mesos.apache.org/documentation/latest/maintenance/) for the cluster node to be removed (via /machine/down). The machine will not be used for new tasks.
+1.  Use the Mesos KillPolicy ([MESOS-4909](https://issues.apache.org/jira/browse/MESOS-4909)) to specify how long, in seconds, [Marathon](https://github.com/mesosphere/marathon/commit/4a6cd7afe66c91741d835a55933da67b72c331de) will wait after sending a SIGTERM before killing the process via SIGKILL.
+1.  Write app to catch SIGHUP and SIGTERM to shutdown gracefully within the killpolicy window. 
+1.  Kill the DC/OS node, wait 21 minutes before looping (to kill another node). DC/OS is configured to rate limit automatic agent removal to 1 every 20 minutes. Do not shrink a cluster faster than this. 
+1.  Wait 1 minutes, then delete the maintenance primitive.  
--- a/1.8/administration/installing/custom/graceful-shutdown.md
+++ b/1.8/administration/installing/custom/graceful-shutdown.md
@@ -3,10 +3,13 @@ post_title: Graceful Node Shutdown
 menu_order: 501
 ---
 
-You can use these instructions to remove a DC/OS cluster node.
+DC/OS supports a partial graceful shutdown of agent nodes. 
+
+You can use these instructions to remove a DC/OS cluster node. 
 
 1.  Add a [Mesos maintenance primitive](http://mesos.apache.org/documentation/latest/maintenance/) for the cluster node to be removed (via /machine/down). The machine will not be used for new tasks.
-1.  Use the Mesos KillPolicy ([MESOS-4909](https://issues.apache.org/jira/browse/MESOS-4909)) to specify how long, in seconds, [Marathon](https://github.com/mesosphere/marathon/commit/4a6cd7afe66c91741d835a55933da67b72c331de) will wait after sending a SIGTERM before killing the process via SIGKILL.
+1.  Use the Mesos [KillPolicy](https://github.com/apache/mesos/blob/1.1.x/include/mesos/mesos.proto#L434-L461) to specify how long, in seconds, [Marathon](https://github.com/mesosphere/marathon/commit/4a6cd7afe66c91741d835a55933da67b72c331de) will wait after sending a SIGTERM before killing the process via SIGKILL.
 1.  Write app to catch SIGHUP and SIGTERM to shutdown gracefully within the killpolicy window. 
 1.  Kill the DC/OS node, wait 21 minutes before looping (to kill another node). DC/OS is configured to rate limit automatic agent removal to 1 every 20 minutes. Do not shrink a cluster faster than this. 
+    **Important:** If you use maintenance primitives, the agent removal rate limit does not apply (`/mesos/master/machine/down`). However, taking a machine down by using the maintenance primitives will also ignore kill policies, and perform force-kills immediately.
 1.  Wait 1 minutes, then delete the maintenance primitive.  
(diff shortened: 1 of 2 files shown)


1.8/administration/installing/custom/graceful-shutdown.md
```diff
@@ -0,0 +1,12 @@
+---
+post_title: Graceful Node Shutdown
+menu_order: 501
+---
+
+You can use these instructions to remove a DC/OS cluster node.
+
+1.  Add a [Mesos maintenance primitive](http://mesos.apache.org/documentation/latest/maintenance/) for the cluster node to be removed (via /machine/down). The machine will not be used for new tasks.
+1.  Use the Mesos KillPolicy ([MESOS-4909](https://issues.apache.org/jira/browse/MESOS-4909)) to specify how long, in seconds, [Marathon](https://github.com/mesosphere/marathon/commit/4a6cd7afe66c91741d835a55933da67b72c331de) will wait after sending a SIGTERM before killing the process via SIGKILL.
```
kaysoky: Instead of the JIRA, try: https://github.com/apache/mesos/blob/1.1.x/include/mesos/mesos.proto#L434-L461

1.8/administration/installing/custom/graceful-shutdown.md
```diff
@@ -0,0 +1,12 @@
...
+1.  Write app to catch SIGHUP and SIGTERM to shutdown gracefully within the killpolicy window. 
+1.  Kill the DC/OS node, wait 21 minutes before looping (to kill another node). DC/OS is configured to rate limit automatic agent removal to 1 every 20 minutes. Do not shrink a cluster faster than this.
```
kaysoky: If you use maintenance primitives, the agent removal rate limit does not apply.  (`/mesos/master/machine/down`)

At the same time, downing a machine with maintenance primitives will also ignore kill policies, and perform force-kills immediately.

1.8/administration/installing/custom/graceful-shutdown.md
```diff
@@ -0,0 +1,12 @@
+---
+post_title: Graceful Node Shutdown
+menu_order: 501
+---
+
+You can use these instructions to remove a DC/OS cluster node.
```
joel-hamill: how about `DC/OS supports a partial graceful shutdown of agent nodes. `

cmaloney:
I don't think users will be able to feasibly use this today. The MAintenance api is hard to use (took me over a day with communication to the people who implemented it to get a simple usage working). It also requires _every_ DC/OS service installed on a cluster to support Maintenance Primitives and have properly set Mesos Kill Polcies, two things we don't enforce in the universe today, and don't do by default inside of marathon apps.

The instructions very in what they do. The first one is a command sent to mesos master directly (which in a "secure" cluster is non-trivial), the second is things that needed to be happening when launching apps for every app on a node (not when trying to gracefully shutdown the node). The third is when writing apps.

The last two are the actual "graceful" shutdown steps. The 21 minute window given isn't correct if a node is _properly_ removed. The proper removal is once all tasks have left the agent as desired run `systemctl kill -s SIGUSR1 dcos-mesos-slave` (See https://github.com/dcos/dcos-docs/blob/6d20498e1c254fec5505191035e572bae6d6a311/1.8/administration/installing/cloud/aws/upgrading.md "deleting agent nodes" or something similar)

To me it feels like the instructions should be broken into two parts. First the "prerequisites" and second how to gracefully shutdown a node.

Prerequisites for getting graceful shutdown: 
 - All Service on the cluster must support Mesos Maintenance Primitives and use KillPolicies
 - All Apps launched via marathon must have a graceful mesos Kill Policy set


Gracefully shutting down a node.
1. Submit a "maintenance window" for the node to mesos (Note: This api is hard to use, produceful no useful error messages when used in ways people didn't think they were using it). https://github.com/dcos/dcos/blob/master/packages/dcos-integration-test/extra/resiliency/test_aws_cf.py#L52 has an example of using the maintenance API. It takes _multiple_ api calls to bring an agent down. First you have to make a "schedule" of the maintenance. That requires that you give a timestamp in nanoseconds (ideally in the future if we want graceful shutdown to work) of when the maintenance window on the agent will start, as well as (potentially) a timestamp when it should end. This needs to have the host identified hte same way it identified itself when registering to mesos (pseudo-standardized in DC/OS but not guaranteed). Once the maintenance schedule is created, then during the maintenance window call `/machine/down` to start the maintenance.
2. Watch for the tasks vacate the agent
3. ssh to the agent, and do a clean agent shutdown using `systemctl kill -s SIGUSR1 dcos-mesos-slave` if it's a private agent or  `systemctl kill -s SIGUSR1 dcos-mesos-public-slave` (See https://github.com/dcos/dcos-docs/blob/6d20498e1c254fec5505191035e572bae6d6a311/1.8/administration/installing/cloud/aws/upgrading.md "deleting agent nodes", might exist in other places)
4. The mesos agent should be fully removed from the cluster. Other agents can be done as desired. Be sure to try and not violate appliaction failure requirements (taking out more replicas of an app than it can handle before it has fully re-replicated).

1.8/administration/installing/custom/graceful-shutdown.md
```diff
@@ -0,0 +1,15 @@
...
+
+You can use these instructions to remove a DC/OS cluster node. 
+
+1.  Add a [Mesos maintenance primitive](http://mesos.apache.org/documentation/latest/maintenance/) for the cluster node to be removed (via /machine/down). The machine will not be used for new tasks.
+1.  Use the Mesos [KillPolicy](https://github.com/apache/mesos/blob/1.1.x/include/mesos/mesos.proto#L434-L461) to specify how long, in seconds, [Marathon](https://github.com/mesosphere/marathon/commit/4a6cd7afe66c91741d835a55933da67b72c331de) will wait after sending a SIGTERM before killing the process via SIGKILL.
+1.  Write app to catch SIGHUP and SIGTERM to shutdown gracefully within the killpolicy window. 
+1.  Kill the DC/OS node, wait 21 minutes before looping (to kill another node). DC/OS is configured to rate limit automatic agent removal to 1 every 20 minutes. Do not shrink a cluster faster than this. 
+    **Important:** If you use maintenance primitives, the agent removal rate limit does not apply (`/mesos/master/machine/down`). However, taking a machine down by using the maintenance primitives will also ignore kill policies, and perform force-kills immediately.
```
cmaloney: /machine/donw doesn't actually kill anything. It just makes it so mesos won't give out any offers.

cmaloney:
If there's a decision to go on with this @jmlvanre needs to review it, he worked a lot on maintenance primitives in mesos

joel-hamill:
@spahl please make the call on this

1.8/administration/installing/custom/graceful-shutdown.md
kaysoky: You mean, the `/maintenance/schedule` does not kill.  `/machine/down` is similar to killing the Mesos agent gracefully, which involves killing all running tasks on that agent.

joel-hamill:
@adam-mesos do you have thoughts on this?

adam-mesos:
I wasn't aware that we were planning to document/recommend the maintenance primitives until we had better framework support and better dcos integration. If you haven't hand testing these instructions, I would be very reluctant to publish them.
I defer to @kaysoky and @jmlvanre for technical details, whenever we are ready to document it.

PR #828, opened 2017-01-04
Fix MesosDNS documentation. 

joerg84:
- [x] Bug fix
- [ ] New feature 

- [x] Blocker
- [ ] High
- [ ] Medium

- [x] I have tested all commands and code snippets.
- [x] I have built locally and tested for broken links and formatting.
- [x] I have made this change for all affected versions (e.g. 1.7, 1.8, and 1.9).
- [x] My doc follows the style guide: https://github.com/dcos/dcos-docs#contributing.

## If you included a comment with your commit, it appears here:

### DIFF
--- a/1.8/usage/service-discovery/mesos-dns/service-naming.md
+++ b/1.8/usage/service-discovery/mesos-dns/service-naming.md
@@ -3,7 +3,7 @@ post_title: Service Naming
 menu_order: 0
 ---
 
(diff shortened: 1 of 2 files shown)


joerg84:
@joel-hamill The issue here is that `$ curl http://master.mesos:8123/v1/enumerate` works fine on OSS clusters, but not on EE clusters due to security where it would require an acs token. This is IMO the easiest fix -keeping it consistent between OSS and EE-, let me know what you think.

PR #912, opened 2017-02-23
Edits the debugging intro to be more descriptive and accurate

emanic:
## Urgency
- [ ] Blocker 
- [x ] High
- [ ] Medium

## Requirements

### DIFF
--- a/1.9/administration/debugging/index.md
+++ b/1.9/administration/debugging/index.md
@@ -4,19 +4,23 @@ feature_maturity: experimental
 menu_order: 3.3
 ---
 
-When things go wrong, you need to see the logs to understand how to fix it. DC/OS services and tasks write `stdout` and `stderr` files in their sandboxes by default. Traditionally, log aggregation has been the solution here. Write the logs locally and then ship all that data elsewhere for someone to actually access. Moving data around is expensive, especially when it ends up being written multiple times.
+The `dcos task exec` DC/OS CLI command allows you to execute commands inside of a task's container. You just need to supply the ID of the task and be logged into the DC/OS CLI via `dcos auth login`. The container must be running. It offers an experience very similar to [`docker exec`](https://docs.docker.com/engine/reference/commandline/exec/), without any need for SSH keys. The connection occurs over an HTTP stream and does not use SSH at all.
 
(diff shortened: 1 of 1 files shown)


1.9/administration/debugging/index.md
```diff
@@ -4,19 +4,30 @@ feature_maturity: experimental
 menu_order: 3.3
 ---
 
-When things go wrong, you need to see the logs to understand how to fix it. DC/OS services and tasks write `stdout` and `stderr` files in their sandboxes by default. Traditionally, log aggregation has been the solution here. Write the logs locally and then ship all that data elsewhere for someone to actually access. Moving data around is expensive, especially when it ends up being written multiple times.
+The `dcos task exec` DC/OS CLI command allows you to execute commands inside of a task's container. All you will need is to be logged into the DC/OS CLI via `dcos auth login`, the ID of the task, and a running container. The `dcos task exec` command offers an experience very similar to [`docker exec`](https://docs.docker.com/engine/reference/commandline/exec/), without any need for SSH keys. The connection occurs over an HTTP stream and does not use SSH at all.
```
klueska: In the first sentence, it sounds a bit weird to reiterate that its a DC/OS CLI command (that should be implied. It might sound better as:
```
The `dcos task exec` command allows you to execute arbitrary commands inside of a task's container.
```

Also, the second sentence seems unnecessary. It adds complicated details about your environment and might be better suited to the section wehre we start talking about actually executing the command.

Finally, the mentioning SSH here seems a bit weird (firstly because docker exec doesn't needs SSH either), but also just the wording --- saying "The connection occurs over an HTTP stream and does not use SSH at all" sounds off somehow.

I would probably reword this whole paragraph to say:
```
The `dcos task exec` command allows you to execute an arbitrary command inside of a task's container and stream its output back to your local terminal. It offers an experience very similar to [`docker exec`](https://docs.docker.com/engine/reference/commandline/exec/). As such, you can execute this command in any of the following four modes:
```

emanic: From what I understand, the fact that you don't need SSH keys is a huge feature. I can see what you mean about going into the protocol. Instinctively, I would agree. At the same time, I have observed people showing great interest in this. Let's drop it but I think we should keep at least the SSH key mention. Thanks @klueska! 👍

PR #1037, opened 2017-05-02
Add API Versioning page

karlkfi:
## Description
- Add API Versioning page
- Fix component header anchor links

Blocking https://github.com/mesosphere/dcos-docs-enterprise/pull/1217 (need to sync this new page to complete `api/index.md`)

## Urgency
- [ ] Blocker 
- [X] High
- [ ] Medium

### DIFF
--- a/1.9/api/index.md
+++ b/1.9/api/index.md
@@ -37,7 +37,7 @@ To determine the URL of your cluster, see [Cluster Access](/docs/1.9/api/access/
 - To determine the full URL of a API resource through a **proxy route**, join the cluster URL, route path, and backend component resource path.
 
     ```bash
-    ${cluster-url}/${route}/${component-resource-path}
+    ${cluster-url}/${route}/${resource-path}
     ```
 
     For example, get the Mesos version from: `https://dcos.example.com/mesos/version`
--- a/1.9/api/index.md
+++ b/1.9/api/index.md
@@ -36,25 +36,25 @@ To determine the URL of your cluster, see [Cluster Access](/docs/1.9/api/access/
 
 - To determine the full URL of a API resource through a **proxy route**, join the cluster URL, route path, and backend component resource path.
 
-    ```bash
-    ${cluster-url}/${route}/${resource-path}
+    ```
+    <cluster-url>/<route>/<resource-path>
     ```
 
     For example, get the Mesos version from: `https://dcos.example.com/mesos/version`
 
 - **File routes** have no backend component, but may serve a directory of files or a single file. So for file routes, specify the file path instead of the backend component resource path.
 
-    ```bash
-    ${cluster-url}/${route}/${file-path}
+    ```
+    <cluster-url>/<route>/<file-path>
     ```
 
     For example, get the DC/OS version of the cluster from: `https://dcos.example.com/dcos-metadata/dcos-version.json`
 
 - **Lua routes** immediately execute code in Admin Router without proxying to an external backend component. So for Lua routes, no path is required after the route.
 
-     ```bash
-     ${cluster-url}/${route}
-     ```
+    ```
+    <cluster-url>/<route>
+    ```
 
      For example, get the public IP of the master node and cluster ID from: `https://dcos.example.com/metadata`
 
--- a/1.9/api/versioning.md
+++ b/1.9/api/versioning.md
@@ -1,27 +1,28 @@
 ---
-post_title: Versioning
+post_title: API Versioning
+nav_title: Versioning
 menu_order: 2
 ---
 
-The DC/OS API is backed by many loosely coupled components, some of which are standalone projects and some of which were designed exclusively for DC/OS. Because of this, several types of versioning mechanisms are supported: component, route, and resource versioning.
+The DC/OS API is backed by many loosely coupled components; some are standalone projects and others are designed exclusively for DC/OS. Because of this, a variety of versioning mechanisms are supported: component, route, and resource versioning.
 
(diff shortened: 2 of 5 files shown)


1.9/api/versioning.md
```diff
@@ -0,0 +1,31 @@
+---
+post_title: Versioning
+menu_order: 2
+---
+
+The DC/OS API is backed by many loosely coupled components, some of which are standalone projects and some of which were designed exclusively for DC/OS. Because of this, several types of versioning mechanisms are supported: component, route, and resource versioning.
```
joel-hamill: `...some are standalone projects and other are designed exclusively for DC/OS. Because of this, a variety of versioning mechanisms are supported: component, route, and resource versioning.`

1.9/api/versioning.md
```diff
@@ -0,0 +1,31 @@
...
+
+To learn how to formulate a specific API call, see the component API reference documentation for that route.
+
+## Component Versioning
```
joel-hamill: If you make these H1s (aka `#`) they will show up on the left hand side "In this post".

1.9/api/versioning.md
```diff
@@ -0,0 +1,31 @@
...
+
+Components with their own open source communities, like Mesos, Marathon, and Mesos DNS, have routes that are based on the well known component name. These routes delegate versioning to the backend component service.
```
joel-hamill: This? `...based on their well known component name.`

1.9/api/versioning.md
```diff
@@ -0,0 +1,31 @@
...
+
+For example, the [Marathon component](/docs/1.9/overview/architecture/components/#marathon) serves the [Marathon API](/docs/1.9/deploying-services/marathon-api/) under the route `/service/marathon` and one of its resource paths is `/v2/apps`, so the full path to that resource is `/service/marathon/v2/apps`.
+
+## Route Versioning
+
+Components that have been specifically designed exclusively for DC/OS tend to follow another versioning pattern, where the name of the component takes a back seat to the name of the feature set. These routes often include a version to make it easier to support renaming or replacing components over time.
```
joel-hamill: - Think you should pick either `exclusively designed for DC/OS` or `specifically designed for DC/OS`, but not both. 
- `tend to follow` - don't they *always* follow another versioning scheme? if not, then suggest the word `...generally follow another versioning...` instead.
-

1.9/api/versioning.md
```diff
@@ -0,0 +1,31 @@
...
+
+For example, the [DC/OS Diagnostics (3DT) component](/docs/1.9/overview/architecture/components/#dcos-diagnostics) serves the [System Health API](/docs/1.9/monitoring/#system-health-http-api-endpoint) under the route `/system/health/v1` and one of its resource paths is `/report`, so the full path to that resource is `/system/health/v1/report`.
+
+## Resource Versioning
+
+Some components avoid path versioning all together and use content negotiation at the resource level to simultaniously support multiple versions of an API at the same path.
```
joel-hamill: `altogether`

1.9/api/versioning.md
```diff
@@ -0,0 +1,31 @@
+---
+post_title: Versioning
+menu_order: 2
```
joel-hamill: should there be a nav_title that says `API Versioning` - wondering if this will help with search.

karlkfi: It'll be under the API Reference section... but i could alternatively have the nav title be "Versioning" and the page title be "API Versioning"...

joel-hamill: yeah, i think adding nav_title would be good - most of our users find things by using search, not TOC - so we have to optimize for what people see with search results.

1.9/api/index.md
```diff
@@ -37,7 +37,7 @@ To determine the URL of your cluster, see [Cluster Access](/docs/1.9/api/access/
 - To determine the full URL of a API resource through a **proxy route**, join the cluster URL, route path, and backend component resource path.
 
     ```bash
-    ${cluster-url}/${route}/${component-resource-path}
+    ${cluster-url}/${route}/${resource-path}
```
sascala: If the curly braces indicate variables, our format is `<dcos-url>/<route>/<component-resource-path>`

1.9/api/versioning.md
sascala: `others are designed`

1.9/api/index.md
karlkfi: `${variable}` is bash string interpolation format. But I can change to `<variable>` if we want to be more consistent.

1.9/api/versioning.md
karlkfi: That... disturbs me. but ok.
Ideally H1's should be for page titles, not top level headings....

1.9/api/versioning.md
karlkfi: > don't they always follow another versioning scheme?
Well, we have some old ones like History Service, which was designed for the Mesosphere Stack before it was DC/OS. It's definitely not well known. There's also Pkgpanda... Basically, anything before we decided that wasn't a sustainable pattern. Unfortunately no one went back and cleaned up the routes afterwards. Its on my TODO list to try to get this kind of thing cleaned up and made more consistent.

karlkfi: I'll use `generally follow another versioning`

1.9/api/versioning.md
karlkfi: sounds good:
```
post_title: API Versioning
nav_title: Versioning
```

PR #1054, opened 2017-05-11
Improve external volumes examples

sascala:
## Description
Have the external volumes examples actually employ volumes
https://jira.mesosphere.com/browse/DOCS-1505

## Urgency
- [X] Blocker 
- [ ] High
- [ ] Medium

## Requirements

### DIFF
--- a/1.9/storage/external-storage.md
+++ b/1.9/storage/external-storage.md
@@ -86,7 +86,7 @@ You can specify an external volume in your Marathon app definition. [Learn more 
       "instances": 1,
       "cpus": 0.1,
       "mem": 32,
-      "cmd": "/usr/bin/tail -f /dev/null",
+      "cmd": "mkdir my-test-vol && touch my-test-vol/test.txt && ls my-test-vol/",
       "container": {
         "type": "MESOS",
         "volumes": [
@@ -136,7 +136,7 @@ Below is a sample app definition that uses a Docker container and specifies an e
       "instances": 1,
       "cpus": 0.1,
       "mem": 32,
-      "cmd": "/usr/bin/tail -f /dev/null",
+      "cmd": "mkdir my-test-vol && touch my-test-vol/test.txt && ls my-test-vol/",
       "container": {
         "type": "DOCKER",
         "docker": {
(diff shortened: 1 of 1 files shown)


1.9/storage/external-storage.md
```diff
@@ -86,7 +86,7 @@ You can specify an external volume in your Marathon app definition. [Learn more
       "instances": 1,
       "cpus": 0.1,
       "mem": 32,
-      "cmd": "/usr/bin/tail -f /dev/null",
+      "cmd": "mkdir my-test-vol && touch my-test-vol/test.txt && ls my-test-vol/",
```
gpaul: I believe this should be 
```
mkdir test-rexray-volume/test-dir && touch test-rexray-volume/test-dir/test.txt && ls test-rexray-volume/test-dir/
```

The `my-test-vol` property that is specified for this volume is the name of the volume in eg., your AWS account, and is unrelated to the directory path in the container. The directory path in the container is given by the `containerPath` property, which in this case is `test-rexray-volume`.

1.9/storage/external-storage.md
```diff
@@ -136,7 +136,7 @@ Below is a sample app definition that uses a Docker container and specifies an e
...
```
gpaul: I believe this should be 
```
mkdir /data/test-rexray-volume/test-dir && touch /data/test-rexray-volume/test-dir/test.txt && ls /data/test-rexray-volume/test-dir/
```

The directory path in the container is given by the `containerPath` property, which in this case is `/data/test-rexray-volume`. Note that it is an absolute path in this case, as this is using the Docker containerizer, which expects absolute paths for `containerPath` and mounts the external volume at exactly the given path in the container.

gpaul (changes requested):
I noticed that the commands are wrong.

If you'd like to test that it is correct now and was wrong before, spin up a cluster on CCM. 

Next, use the following CMD instead.

```
"cmd": "date >> test-rexray-volume/test.txt; cat test-rexray-volume/test.txt",
```
(It's important to use the '>>' or "append" operator for this test.)

If the external volume is being used correctly, the logs of successive runs of the application will show more and more lines as part of the `cat` operation.

Finally, submit the app definition and wait for it to run - it should run several times and every time the list of dates printed to the stdout log of the instances should grow longer. If it doesn't it means that the `test.txt` file is not being persisted between runs and the config is wrong.

The same test holds for the docker containerizer, but the paths should be `/data/test-rexray-volume/test.txt` instead.

sascala:
Thank you for the suggested command, @gpaul . I've changed the command and added explanatory text.

gpaul (approved):
LGTM :+1:

PR #1113, opened 2017-06-23
Pods Updates

sascala:
## Description
https://jira.mesosphere.com/browse/DOCS-1778

## Urgency
- [ ] Blocker 
- [ ] High
- [X] Medium

## Requirements


sascala:
@kensipe Thank you for your feedback in the Jira. I've incorporated your comments here and, if the changes look good to you, I'll port this whole document over to Marathon. I had a few questions, which I'll post here.

- You wrote, "networks have names and that they are specific to your env." Can you tell me more about that? I understand that some other part of the environment needs to recognize and reference the name. But, just to allay some of my networking ignorance, which parts?

- Do non-dcos setups also have default virtual network names? (important when I port this doc to the marathon site).

- If you have more than one virtual network (if you can?), then you've got to at least specify a second name, right?

- You wrote, "When dealing with virtual networks... it means that the host network and IP address scheme is NOT used." But when I ssh to the master node of my cluster, not only can I ping `9.0.2.2`, I can also ping `10.0.6.26` (an IP taken from the output of `dcos node`). So, is it that we can ping both the actual node IPs _as well as_ the virtual network IPs we've created?

- I noticed I _can't_ ping the virtual network by name (i.e., `ping dcos` doesn't work from the master node). Where is this network name used? I'm assuming something somewhere translates the name to the appropriate IP, but that something isn't functioning on the master node?

kensipe:
@sascala 

1.  if you go to dcos_url/`#/networking/networks` you will see a list of network names.  They are the only names which are allowed.   These are configured as part of installation.    The default is `dcos`.  The way to think of these networks is they are separate.   you have the default cluster network (and it's IP addresses) and you have the `dcos` network and it's separate IP addresses.  It is possible that during installation another named network is provided with it's network.   This is a way for services on dcos to separate their network traffic.   The networking team is best to explain any setup details.

1. non-dcos setups for networks... that's a great question.   It is something that is configured as part of Mesos as I understand it.   This would make it possible for someone setting up Mesos for themselves to provide a network.   If true, it makes the answer to your question yes.   We may want to check with product first, then core.  @greggomann Do you know?

1. re: virtual networks... I don't know if I understand your question... it is definitely possible to have multiple networks.   Each services can use the named virtual network they prefer.   It is also suppose to be possible to have an app use multiple virtual networks.  In this case the second named virtual name would need to be provided.  I have not tested this...  We do not have test environments for this that I'm aware of.

1. re: pinging and IPs.   It is possible to have a vNet and be able to ping in the way you describe.. however this shouldn't be possible in "container/bridge" mode.  

1. the name "dcos" for the network is translated as you mentioned.   I'm not sure the exactly what makes the translation.   There is definitely a component that would need to track and provide housekeeping on IP addresses (like a service needs 1 and it can't be reused).    This is outside of Marathon.     I was assuming that the executor plays apart... but there needs to be something centralized that tracks.

1. I'm not sure on the mulit-pod... where did it come from?  if you need me to.. I can check on it tomorrow.

1.9/deploying-services/pods/technical-overview.md
```diff
@@ -17,7 +17,7 @@ You configure a pod via a pod definition, which is similar to a Marathon applica
 # Networking
 Marathon pods only support the [DC/OS Universal container runtime](/docs/1.9/deploying-services/containerizers/), which supports multiple image formats, including Docker.
 
-The Universal container runtime simplifies networking by allowing the containers of each pod instance to share a network namespace and communicate over localhost. If you specify a container network without a name in a pod definition, it will be assigned to the default network.
+The Universal container runtime simplifies networking by allowing the containers of each pod instance to share a network namespace and communicate over localhost. If you specify a container network without a name in a pod definition, it will be assigned to the default network. If you have installed DC/OS [our AWS templates](/docs/1.9/installing/cloud/aws/), the default network is `dcos`.
```
kensipe: I do not like the language here..  `localhost` is not the right word to use.   they will communicate over a private network. or a vlan.

sascala:
@joel-hamill Once you've looked this over, I'll copy it to 1.10.

1.9/deploying-services/pods/technical-overview.md
```diff
@@ -151,6 +151,7 @@ The following JSON specifies a Docker image for a pod:
 
 # Limitations
 
+- Pods are not supported in DC/OS Enterprise `strict` security mode.
```
joel-hamill: `Enterprise DC/OS`, not the other way around. Also, no need to have strict in command syntax - it would probably be more useful if you made it all a hyperlink to the definition instead, e.g.: `Enterprise DC/OS [strict security mode](docs.mesosphere.com/security/..).` .

sascala:
@joel-hamill I addressed your final comment

sascala:
...and I updated the 1.10 pods docs as well.

1.9/deploying-services/pods/technical-overview.md
```diff
@@ -17,7 +17,7 @@ You configure a pod via a pod definition, which is similar to a Marathon applica
 # Networking
 Marathon pods only support the [DC/OS Universal container runtime](/docs/1.9/deploying-services/containerizers/), which supports multiple image formats, including Docker.
 
-The Universal container runtime simplifies networking by allowing the containers of each pod instance to share a network namespace and communicate over localhost. If you specify a container network without a name in a pod definition, it will be assigned to the default network.
+The Universal container runtime simplifies networking by allowing the containers of each pod instance to share a network namespace and communicate over a VLAN or private network. If you specify a container network without a name in a pod definition, it will be assigned to the default network. If you have installed DC/OS using [our AWS templates](/docs/1.9/installing/cloud/aws/), the default network is `dcos`.
```
joel-hamill: probably don't need `...our AWS templates...`, rather `...the AWS templates...`

sascala:
@joel-hamill fixed

joel-hamill:
@sascala should i merge?

sascala:
Yes please!

PR #1228, opened 2017-09-26
Update Components for 1.10

karlkfi:
## Description
- AdminRouter now does dynamic DNS resolution and no longer requires an external reloader process
- pkgpanda socket is managed by gunicorn instead of systemd now to avoid a race condition during upgrades
- I opted not to remove minuteman as a component, even tho it no longer shows up in the component list, because it is still a separate application even if it's not a systemd service. It's running in an erlang vm shared with navstar. In 1.11 more components will share the same erlang VM and they will also have seperate systemd watchdogs. So Minuteman Watchdog will show up again in the list in 1.11. Removing and re-adding it would be confusing to users.

## Urgency
- [ ] Blocker 
- [X] High
- [ ] Medium


1.10/overview/architecture/components.md
```diff
@@ -648,58 +640,94 @@ To see a list of the systemd components running on any particular node, list the
...
+dcos-pkgpanda-api.service
+dcos-rexray.service
+dcos-signal.timer
+dcos-spartan.service
+dcos-spartan-watchdog.service
+dcos-spartan-watchdog.timer
 ```
+
+
+# Changes Since DC/OS 1.9
```
stbof: Please apply same suggestions as enterprise version.

